Issue #1, opened 2020-07-01
polymorphic suggestion for unhandled cases in switch uses backtick


Coobaha:
```
let withpayload = #polymorphic(3)
switch withpayload {
  | #polymoprhic(c) => Js.log(c)
}
```

```
 13 │ 
  14 │ let withpayload = #polymorphic(3)
  15 │ switch withpayload {
  16 │   | #polymoprhic(c) => Js.log(c)
  17 │ }
  18 │ 
  19 │ let array = [1,2,3]
  
  You forgot to handle a possible case here, for example: 
`polymorphic _
```

Coobaha:
I should have opened this in bs-platform directly maybe.. cc @bobzhang

chenglou:
Just to make sure: you saw that you made a typo right?
And yeah this isn't syntax-related =). Thanks. Closing!

Coobaha:
Oh yeah 😅  error suggested case uses ``` instead of `#

chenglou:
Whoops my mistake. That's a printing error.

cristianoc:
Could be, this is about warnings vs errors. Unless there's another repro with errors.
CC @bobzhang

Coobaha:
Another old issue with suggestions that exists in `.re` is that suggestions are printed in `.ml` syntax.
 
```
type myT =
  | A
  | B
  | C
  | D(int)

switch (A) {
  | A => 1
}

```

```
  26 │ switch (A) {
  27 │   | A => 1
  28 │ }
  29 │ 
  30 │ let array = [1,2,3]
  
  You forgot to handle a possible case here, for example: 
(B|C|D _)
```

bobzhang:
This is a long existing issue (with refmt too), seems we need to patch 
`val Parmatch.pretty_val : Format.formatter -> Typedtree.pattern -> unit`
cc @cristianoc @IwanKaramazow

bobzhang:
Note there's an easy solution:

We can convert Typedtree.pattern to Parsetree.pattern
pat: mapper -> Typedtree.pattern -> pattern;
So that if res gives us a printer for Parsetree.pattern, it should work

PR #12, opened 2020-07-03
Add `if let` syntax

tomgasson:
`if let` brings pattern matching to `if`
---------------------------------------
`if let PAT = EXPR { BODY }`

The motivation for having any construct at all for this is to simplify the cases that today call for
a switch statement with a single non-trivial case. This is predominately used for unwrapping
`option<'t>` values, but can be used elsewhere.

Before:
```reason
switch(optVal) {
    | Some(x) => doSomethingWith(x)
    | _ => ()
}
```

After:
```reason
if let Some(x) = optVal {
    doSomethingWith(x)
}
```

Complex example:
```reason
if let Some(x) = optVal when x > 10 {
    doSomethingWith(x)
} else if let Theme({color: Red | Blue, emoji}) = getTheme() {
    doSomething()
} else {
    doSomethingElse()
}
```

There are >500 occurrences in FB of `| _ => ()` where this form would be more appropriate

This is syntax sugar only, it's implemented similar to `ns.ternary` so that that it should be stable across parsing and printing

Prior Art
---------
https://github.com/facebook/reason/issues/1301

* Rust - https://github.com/rust-lang/rfcs/pull/160. The same justifications apply here

* Swift `if case x {}` and `if var = foo()` (for reference types)

* Clojure https://clojuredocs.org/clojure.core/if-let

### DIFF
--- a/src/napkin_printer.ml
+++ b/src/napkin_printer.ml
@@ -2365,6 +2365,58 @@ and printExpressionWithComments expr cmtTbl =
   let doc = printExpression expr cmtTbl in
   printComments doc cmtTbl expr.Parsetree.pexp_loc
 
+and printIfChain pexp_attributes ifs elseExpr cmtTbl =
+  let ifDocs = Doc.join ~sep:Doc.space (
+    List.mapi (fun i (ifExpr, thenExpr) ->
+      let ifTxt = if i > 0 then Doc.text "else if " else  Doc.text "if " in
+      match ifExpr with
+        | ParsetreeViewer.If ifExpr ->
+          let condition =
+            if ParsetreeViewer.isBlockExpr ifExpr then
+              printExpressionBlock ~braces:true ifExpr cmtTbl
+            else
+              let doc = printExpressionWithComments ifExpr cmtTbl in
+              match Parens.expr ifExpr with
+              | Parens.Parenthesized -> addParens doc
+              | Braced braces -> printBraces doc ifExpr braces
+              | Nothing -> Doc.ifBreaks (addParens doc) doc
+          in
+          Doc.concat [
+            ifTxt;
+            Doc.group (condition);
+            Doc.space;
+            let thenExpr = match ParsetreeViewer.processBracesAttr thenExpr with
+            (* This case only happens when coming from Reason, we strip braces *)
+            | (Some _, expr) -> expr
+            | _ -> thenExpr
+            in
+            printExpressionBlock ~braces:true thenExpr cmtTbl;
+          ]
+        | IfLet (pattern, conditionExpr) ->
+          Doc.concat [
+            ifTxt;
+            Doc.text "let ";
+            printPattern pattern cmtTbl;
+            Doc.text " = ";
+            printExpressionWithComments conditionExpr cmtTbl;
+            Doc.space;
+            printExpressionBlock ~braces:true thenExpr cmtTbl;
+          ]
+    ) ifs
+  ) in
+  let elseDoc = match elseExpr with
+  | None -> Doc.nil
+  | Some expr -> Doc.concat [
+      Doc.text " else ";
+      printExpressionBlock ~braces:true expr cmtTbl;
+    ]
+  in
+  Doc.concat [
+    printAttributes pexp_attributes cmtTbl;
+    ifDocs;
+    elseDoc;
+  ]
+
 and printExpression (e : Parsetree.expression) cmtTbl =
   let printedExpression = match e.pexp_desc with
   | Parsetree.Pexp_constant c -> printConstant c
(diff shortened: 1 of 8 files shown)


IwanKaramazow:
Good stuff! Looks good at first sight, I'll do a more thorough review during the weekend.

bloodyowl:
I like the idea, but feel like the legibility is pretty hard.

I'd find it clearer to read something like the following: 

```re
if optVal match Some(x) {
  doSomethingWith(x)
}
```

What do you think?

tomgasson:
@bloodyowl 
Rust also discussed `optVal is Some(x)` in a similar form to your `optVal match Some(x)`  suggestion, and I think the same reasoning would apply here https://github.com/rust-lang/rfcs/pull/160#issuecomment-48553528

 - `let` is already a keyword an this is just a new position to use it
 - `let` disambiguates the parsing immediately after `if`
 - The ordering is already equivalent with let bindings `let MyVal(x) = foo()`

tomgasson:
> Good stuff! Looks good at first sight, I'll do a more thorough review during the weekend.

Thanks - I haven't paid too much attention to breadcrumbs/regions, exception cases or doc printing so could probably use some guidance on how to do them best

bloodyowl:
@tomgasson Good points yeah. I guess that's where the most sensible tradeoffs are made.

chenglou:
So to be clear, no `when` support nor let chaining for now right? These can be postponed

I'm still torn between `if let Some(a) = result` and `if let a = result`. I'm not sure the generality of the former is worth it. I'm also not sure the latter covers enough.

cristianoc:
Wondering how nested optional record fields look like.

Before:
```reason
type point = { x:int, y:int, z:option<int> }

type nested = { origin:option<point>, name:string }

let getZ = (nested) => switch nested.origin {
  | None => 0
  | Some(point) => switch point.z {
    | None => 0
    | Some(z) => z
  }
}
```

after:
```reason
let getZ = nested =>
  if let Some(point) = nested.origin  {
    if let Some(z) = point.z  {
      z
    } else {
      0
    }
  } else {
    0
  }
```

@tomgasson notices there's an extra space before`{` on the second line.

src/napkin_printer.ml
```diff
@@ -2365,6 +2365,70 @@ and printExpressionWithComments expr cmtTbl =
...
+          let patternDoc = printPattern pattern cmtTbl in
+          let conditionDoc = printExpressionWithComments conditionExpr cmtTbl in
+          let guardDocs = match guard with
+          | Some expr -> Doc.concat [
+            Doc.text "when ";
+            printExpressionWithComments expr cmtTbl;
+          ]
+          | None -> Doc.nil
+          in
+          Doc.concat [
```
IwanKaramazow, line 2405: Can you add an extra test case for what happens with `if let Some(suuuuuuuuuuuuuuuuuuuuuuuuuperLongName) = anotherSuuuuuuuuuuuuuuuuuuuuuuuuuuuperLongName() {` ?

One possible option for a layout is having a newline after the `if let Some(x) =` when things exceed the line width.
```ocaml
Doc.group ( (* make this a separate group *)
  Doc.concat [ (* list of things *)
    lhs;
    Doc.text " =";
    Doc.indent ( (* increase the indentation *)
      Doc.concat [ (* list of things *)
        Doc.line; (* if the layout fits, make this a space, if it doesn't fit, make it a newline *)
        rhs
      ]
    )
  ]
)
```

IwanKaramazow: Can we maybe also add a test case for a record destructure that exceeds the line width:
```
if let {
  suuuuuuuuuuuuuuuuuuuuuuuuuuuperLongName,
  suuuuuuuuuuuuuuuuuuuuuuuuuuuperLongName2,
  suuuuuuuuuuuuuuuuuuuuuuuuuuuperLongName3,
} = buildMyRecord() {
```

IwanKaramazow: Feel free to ping me if you need help.

yawaramin:
Does this PR suppress the fragile pattern match warning (4)? I guess it wouldn't, correct?

src/napkin_printer.ml
tomgasson: Added tests. The grouping is already ideal IMO because the pattern and the expression already break themselves down if they don't fit

tomgasson:
@yawaramin: If there's an ` else { expr }` it's encoded as `| _ => expr` and if not there's an implicit `| _ => ()` used, so this will never trigger that warning

tests/printer/expr/__snapshots__/render.spec.js.snap
```diff
@@ -2198,6 +2198,195 @@ let x =
...
+
+// Ternary
+let a = b
+  ? if let Some(x) = foo() {
+      1
+    } else {
+      2
+    }
+  : 3
+
```
tomgasson, line 2370: This could alternately print as
```reason
let a = b ? if let Some(x) = foo() {
  1
} else {
  2
} : 3
```
But that would be a change to ternary printing, not `if let`

yawaramin:
@tomgasson thanks. What I'm thinking of is something like:

```reason
let x = if let Ok(result) { 1 } else { 0 }
```

...desugaring to:

```reason
let x = switch result {
  | Ok(_) => 1
  | _ => 0
}
```

...which indeed seems to trigger the warning.

tomgasson:
@yawaramin: Ah, I see. Yeah I think we should then suppress it then, since the intention here is to allow that

but you've raised a few more cases of interest to me

```
let x = if let Ok(_) = result {
  1
} else if let Error(_) = result {
  2
} else {
  0
}
```
...the `else` will never be taken. `else if` becomes a nested switch rather than lifting the conditions up to the same level because that's better in the general case (if it was `getResult()` and side-effecting or if they were different `result` and `other`)

```
let x = if let Ok(_) | Error(_) = result {
  1
} else {
  0
}
```
...the else will never be take. The existing warning should let us know

tests/printer/expr/__snapshots__/render.spec.js.snap
IwanKaramazow: Don't worry about this, it's not worth the time right now.

yawaramin:
@tomgasson yeah, in that case we would hit warning 11 (this match case is unused). But just to clarify, you do want to suppress warning 4 for this construct.

Another question, will warning 11 wording also be adjusted to reflect the sugar syntax? I.e. it would be confusing to see `this match case is unused` for an unused `else` clause.

tomgasson:
Updated to suppress warning 4.

For the wording of 11, I think we need the capability to rewrite warnings as mentioned in #1

IwanKaramazow:
@tomgasson Merging, thanks for the implementation. Nice work!

PR #19, opened 2020-07-05
Token.toString Lident and Uident

chenglou:
Was printing `list`. It fell into Lident in my change but toString showed `list` so that was confusing


chenglou:
Should I handle other ones too, e.g. `String`?

PR #21, opened 2020-07-06
Suport for `list{1,2,3}` instead of `list[1,2,3]`.

cristianoc:
Closes https://github.com/BuckleScript/syntax/issues/8


cristianoc:
A few examples can be added, such as

```reason
let x = <div> list{123} list {123} </div>
```

PR #27, opened 2020-07-07
Expose print interface

IwanKaramazow:


src/napkin_multi_printer.mli
```diff
@@ -0,0 +1 @@
+val print: [ (* `refmt of string (* path to refmt *) | `ml *) `res ] -> input: string -> unit
```
bobzhang, line 1: assuming this is the public interface, the return value should be string?

IwanKaramazow: still WIP, typo, should be string

bobzhang: or it should be 
```
val print : ... -> input:string -> Format.formatter -> unit
```
whichever you like

PR #31, opened 2020-07-08
Make `catch` not a keyword, but just an identifier.

cristianoc:
Fixes https://github.com/BuckleScript/syntax/issues/28


cristianoc:
Still need to tweak the error message.

PR #33, opened 2020-07-08
Add functions accessing diagnostics startPos / endPos

### LINKED ISSUE #32: napkin_diagnostics: Give access to start / end pos
As far as I understand, a failing parse can yield multiple diagnosis records, each with their own start / end pos.
`Napkin_diagnostics.t` [doesn't expose the record internals](https://github.com/BuckleScript/syntax/blob/master/src/napkin_diagnostics.mli#L3), so I have no idea how to access them.

I guess we could expose this information with some helper functions?
The reason why I need this is that I need the line, column start and column end for each error to display in the playground editor and report the location correctly

ryyppy:
Fixes #32

### DIFF
--- a/src/napkin_diagnostics.ml
+++ b/src/napkin_diagnostics.ml
@@ -30,6 +30,9 @@ let parseReportStyle txt = match (String.lowercase_ascii txt) with
   | "plain" -> Plain
   | _ -> Pretty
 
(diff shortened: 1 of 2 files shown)


IwanKaramazow:
Thanks, perfect!

PR #37, opened 2020-07-10
Expose Napkin_diagnostic's toString and explain function

ryyppy:
Exposes the Diagnostic `toString` and `explain` function as public API.

## Why?

1) The playground needs a way to transform a single `Diagnostic.t` into a pretty report string (right now I can only stringify a list of `Diagnostic.t` with `stringOfReport`)
2) The playground needs a small msg string for hover message display (basically only the thing `explain` returns)


IwanKaramazow:
If you need other formats or diagnostics data, let me know.  The sky is the limit.

PR #39, opened 2020-07-11
Mark if let as experimental

IwanKaramazow:
```
if let Error(x) = result {
  Js.log("The sky is red")
} else if let Ok(y) = result {
  Js.log("The sky is blue")
} else {
  Js.log("no sky")
}
```
image


chenglou:
Yeah this needs a bit more caution. Since it's additive we can postpone it until we work out the stance on the compiler warnings, the actual behavior of `else if let`, how to prevent abuse of this syntax, etc.

chenglou:
Also this flexes the error messages system so much. This is great

PR #51, opened 2020-07-16
Make location of variant cases align with the id, not `|`.

cristianoc:
See https://github.com/BuckleScript/syntax/issues/50


cristianoc:
nope

PR #64, opened 2020-07-21
Pun record declaration fields when possible.

IwanKaramazow:
```
type a = string
type b = int

type r = {a, b}
```

Requested in https://github.com/BuckleScript/syntax/issues/47


cristianoc:
@IwanKaramazow was there something to be careful about for unary records?

IwanKaramazow:
That was in record expressions. Not yet sure how we could tackle that one

chenglou:
Let's hold off on this one. Not sure about printing record punning yet

cknitt:
@chenglou What exactly are you unsure about / what are your concerns?

chenglou:
cknitt it's possible that we reconsider punning in favor of some other syntax possibilities

bobzhang:
If we allow punning in writing, we should do punning in printing, 
otherwise we should not allow punning in writing either. 

This would create confusion for example, people write code:
```reasonml
type t = { hi, lo}
```
and formatted as 
```reasonml
type t = { hi : hi, lo : lo} 
```
People will read it that you are against such style, if you are against such style, why not prohibit in the beginning (in the parsing)?.

I am not against/for punning, but we should be consistent here

cknitt:
@chenglou I agree with @bobzhang that we should be consistent here.

Also, I think it would be a mistake to remove punning support because:

1. It would make code more verbose/less readable.
1. It was supported in the old Reason parser.
1. Why support it for named function arguments, but not for record fields?
1. I thought the idea for Reason syntax was to be familiar to JS developers, and JS supports punning for objects fields, too.

In short, not supporting punning would feel like a regression to me, both compared to the old Reason parser and to JS.

So I am quite curious about what the "other syntax possibilities" are that outweigh these concerns.

bobzhang:
@cknitt I digged into a little bit. The proposal is to remove punning in type declarations not in expressions where punning is mostly used. Does it sound good to you?

cknitt:
@bobzhang Thanks for digging! That's good news. I had understood Cheng's comments to mean that he would like to remove punning altogether.

> @cknitt I digged into a little bit. The proposal is to remove punning in type declarations not in expressions where punning is mostly used. Does it sound good to you?

pckilgore:
I don't mind losing punning on types  fields (data fields a different story) but agree with Bob that I would dislike the inconsistency of considering it valid syntax but rewriting it on fmt.

ryyppy:
@IwanKaramazow I guess the conclusion of this discussion is that we want to keep the record field declarations always explicit. For punning record values, we already merged #480.

So I'd propose to close this for now.

PR #75, opened 2020-07-28
temporary solution: adding locations for error reporting

bobzhang:
cc @chenglou @IwanKaramazow 
This is a temporary solution for adding locations in error reporting. Feel free to remove such piece of code when you finish the `exit` approach


chenglou:
Will be superseded by #77

Issue #68, opened 2020-07-25
Error reporting format (and bsc/bsb integration)


chenglou:
This is probably highest priority. We need to fix this. cc @IwanKaramazow @bobzhang 

1. This is the reporting from the syntax itself:
<img width="1003" alt="Screen Shot 2020-07-25 at 4 18 15 PM" src="https://user-images.githubusercontent.com/1909539/88468003-731c4100-ce92-11ea-88a0-99dc3c5e51a1.png">

2. This is from `bsc -fmt` (no super errors):
<img width="878" alt="Screen Shot 2020-07-25 at 4 19 14 PM" src="https://user-images.githubusercontent.com/1909539/88468014-97781d80-ce92-11ea-8569-872985b58f5d.png">

3. And this is from `bsb` + super errors (the non-super-errors version looks the same as the previous screenshot):
<img width="995" alt="Screen Shot 2020-07-25 at 4 29 36 PM" src="https://user-images.githubusercontent.com/1909539/88468132-09049b80-ce94-11ea-829d-90b30644747a.png">

Issues with 1:
We should probably drop the display now since this'll mainly be used in bsc, aka should go through the normal or super errors reporting pipeline (aka should not make its own inconsistent error display). Regarding the error location formatting, I saw https://github.com/BuckleScript/syntax/blob/a612f224a0f08e55b7f016cf7c9c4ae533f7e2a4/src/napkin_diagnostics.ml#L164
But that isn't needed if RLS turns on the env var `BS_VSCODE=1`: https://github.com/BuckleScript/bucklescript/blob/81b141146f520c804fb6a4fbb6f8df2d38522860/jscomp/super_errors/super_main.ml#L3
(Honestly I don't like that env var. And with the upcoming language-server changes, it also won't be needed. Also that env var disables super errors' much friendlier error explanations. Kinda like throwing the baby out with the bathwater here.)

Issues with 2:
New syntax files have super-errors turned on. `bsc -fmt` doesn't. This is inconsistent. Notice also the double location report. First location is `reason-react/_none_`??

Issues with 3:
Reporting has no indentation and double location like in previous issue (this isn't related to super-errors). If we solve the previous 2 issues, these ones will naturally go away.

chenglou:
Fixed by #87

PR #77, opened 2020-07-28
Exit parser instead of raising error

chenglou:
Exit parser instead of raising error

See #68. The problem we have is the 2nd and 3rd screenshots, with a superfluous and nonexistent location reported.

This happens because BS/OCaml's way of reporting an error is to catch e.g. a `Location.Error` exception. The reporting displays the location, then move on. This is problematic for our syntax, because we report more than one error (i.e. we have more than one location to show).

The theoretical solution is to refactor the BS/OCaml reporting pipeline to accept a new exception that allows more than one error. But this shakes things up too much, so our temporary solution is to print all the errors on the syntax side, then simply exit the process. No exception raised means the BS side also doesn't catch anything and tries to display that nonsensical location line.

Caveats
- (Not implemented yet) We need to reproduce the display logic inside the syntax code. We'll use the same display logic as BS' super-errors for now, and assume that most folks using BS also have super-errors turned on (which is the default). The result is that when we run bsb, a syntax error display and a compiler error display are visually consistent, even though they're produced by 2 different pieces of code. The look will be inconsistent if user has super-errors disabled; a type error (from BS) would look rustic, while a syntax error (from here) still looks super. Technically the compiler can invoke the syntax api and pass a flag to make it display the old ocaml error format, but nah.
- Exiting a process is a hard control flow. The compiler can't even "catch" it. This is problematic if the compiler wants to keep operating on the file after syntax errors. Right now there's no such thing as "catching a syntax error and recover and keep analyzing the file", but we'll likely need this in the future when in the editor, the compiler operates on the recovered syntax tree in order to provide best-effort type hint, autocomplete, etc. Can't have the syntax exit the compiler process there =).
- The online playground, compiled through jsoo, naturally doesn't understand "exiting a process". We'll need some patching here.

Anyway, temporary, rather stable solution. After other things settle a bit, we'll come back to this and properly fix it.

### DIFF
--- a/src/napkin_driver.ml
+++ b/src/napkin_driver.ml
@@ -95,10 +95,10 @@ let parse_implementation sourcefile =
   if parseResult.invalid then begin
     let style = Napkin_diagnostics.parseReportStyle "" in
     let msg = Napkin_diagnostics.stringOfReport ~style parseResult.diagnostics parseResult.source in
-    raise (Location.Error (Location.error msg))
+    print_endline msg;
+    exit 1
   end;
   parseResult.parsetree
-  [@@raises Location.Error]
 
 let parse_interface sourcefile =
   Location.input_name := sourcefile;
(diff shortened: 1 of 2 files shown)


chenglou:
cc @IwanKaramazow @bobzhang @ryyppy

ryyppy:
I think this will actually not interfere with the JSOO implementation since we are overriding the Napkin_driver functionality and using bits of the Napkin_diagnostics module directly

Issue #97, opened 2020-08-09
Give good error when a type declaration has an inline record type when not permitted


chenglou:
<img width="304" alt="image" src="https://user-images.githubusercontent.com/1909539/89730323-694a2000-d9f2-11ea-95ca-822e17e031ae.png">

PR #114, opened 2020-08-19
Give good error when a type declaration has an inline record type when not permitted

IwanKaramazow:
Fixes https://github.com/rescript-lang/syntax/issues/97


sikanhe:
Hey - is it better to say something like: "Nested record types must be defined separately"? And perhaps give a simple example?

chenglou:
Nested isn't as accurate as inline. And when we talk about defined separately we'd have to exclude variant constructors, which comes back to this

PR #118, opened 2020-08-26
escape LineTerminator in escape sequences template literal characters

IwanKaramazow:
Interprets escape sequences in template characters as specified in:
https://tc39.es/ecma262/#prod-TemplateCharacter

The "LineTerminator" is part of "NonEscapeCharacter" in a "EscapeSequence"

Example:
```
`This is a long string with no\
carriage return`
```
related: https://forum.rescript-lang.org/t/syntax-to-break-long-single-line-strings


IwanKaramazow:
If you want to use line breaks to make your source code easier to read, but you don’t want the line breaks to be part of the string’s value, write a backslash (\) at the end of those line.

PR #125, opened 2020-08-30
Parse <Foo.bar /> elements

jchavarri:
Found in #124.

Reason parser used to parse JSX elements like `<Foo.bar />`:
https://astexplorer.net/#/gist/e21abaf96358b68be3ed3da19930be58/32e083720817c639502960b54984dff8396a3f7b.

React ppx also builds on top of this to allow custom `make` names, e.g.
```reason
let t = <Foo.bar />
let t = <Foo.Bar />
```
is transformed into:
```reason
let t = React.createElement(Foo.bar, Foo.barProps()); /* We can call `bar` instead of `make` */
let t = React.createElement(Foo.Bar.make, Foo.Bar.makeProps());
```

(shameless plug, can be tested in https://jchavarri.github.io/ppx-explorer/ 😜 )

### DIFF
--- a/tests/parsing/grammar/expressions/jsx.js
+++ b/tests/parsing/grammar/expressions/jsx.js
@@ -18,9 +18,24 @@ let _ = <div className="menu"></div>
 let _ = <div className="menu" onClick={_ => Js.log("click")}> </div>
 let _ = <div className="menu" onClick={_ => Js.log("click")}></div>
 
+let _ = <Navbar />
 let _ = <Navbar> </Navbar>
-let _ = <Nav.Navbar> </Nav.Navbar>
-let _ = <Nav.navbar> </Nav.navbar>
+let _ = <Navbar></Navbar>
+let _ = <Navbar className="menu"> </Navbar>
+
+let _ = <Dot.Up />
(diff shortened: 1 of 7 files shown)


tests/parsing/grammar/expressions/jsx.js
```diff
@@ -20,6 +20,7 @@ let _ = <div className="menu" onClick={_ => Js.log("click")}></div>
 
 let _ = <Navbar> </Navbar>
 let _ = <Nav.Navbar> </Nav.Navbar>
+let _ = <Nav.navbar> </Nav.navbar>
```
IwanKaramazow, line 23: Can you add more test cases? Nested, self closing etc.
Can you also test cases for the printer in `tests/printer/expr/jsx.js`?
Can you also add a test case in `tests/parsing/errors/expression/jsx` with an error in the closing "name"?
`<Nav.navbar> </Nav.foo>`?

jchavarri: Thanks @IwanKaramazow, added in 6987427.

IwanKaramazow:
@jchavarri Muchas gracias!

PR #128, opened 2020-09-02
Add esy

Et7f3:
This PR reduce the gaps between native/rescript by allowing native user to use the same syntax. It is very low overhead. Ideally you have to bump the version number on update. or because this project is "private": true we will have to consume with github hash so version number is useless and no overhead.
I can't expose benchmarkings since it use macos specific code.
Since rescript copy rescript files during his compilation process I haven't set up `esy release` that allow to install only the rescript binary with npm.


.gitattributes
```diff
@@ -0,0 +1,12 @@
...
+*.re text eol=lf
+*.rei text eol=lf
+*.res text eol=lf
+*.resi text eol=lf
+
+# Recognize all re files as Reason
+*.re linguist-language=Reason
+
+# Recognize all ml files as OCaml
+*.re linguist-language=OCaml
```
eWert-Online, line 12: Wouldn't this have to be *.ml ?

Et7f3: yes lol (copy paste error)

chenglou:
Got stale and we've decided to go to a different direction as you probably know, sorry.

Thanks for the contribution though

PR #132, opened 2020-09-05
Provide specific error for uppercase typedef

### LINKED ISSUE #122: Provide good error message when the type is capitalized
Hot newcomer path probably.

jchavarri:
Fixes #122. It works, but it is probably still far to be the right fix 😅 Any feedback is appreciated.

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -493,8 +493,9 @@ let parseHashIdent ~startPos p =
   parseIdent ~startPos ~msg:ErrorMessages.variantIdent p
 
 (* Ldot (Ldot (Lident "Foo", "Bar"), "baz") *)
-let parseValuePath p =
+let parseValuePath ?customError p =
   let startPos = p.Parser.startPos in
+  let endPos = p.Parser.endPos in
   let rec aux p path =
     match p.Parser.token with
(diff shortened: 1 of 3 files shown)


IwanKaramazow:
Looking into this PR now. I'm thinking about ways to handle this in a more generic way.

IwanKaramazow:
After having given it some thought, this looks like the correct solution. I was thinking that we should solve it in a more “generic” way, but that turned out to be magic. One of the goals of the new syntax is to keep the implementation of the parser as simple as possible. Your intuition was good. Thanks for helping out here.

A couple of suggestions:
1. Can we change the parameter `?customError` into `~msg`? It would be good if the caller of `parseValuePath` needed to pass an error message. This way we “force” good error messages everywhere =D Can we pass an appropriate error message everywhere? Should this just be a simple string?
2. Would it be possible to add test cases for every possible situation where `parseValuePath` is called with an uppercased ident? We can dump them all in one file `tests/parsing/errors/other/` if you want.
3. Unrelated, but should we rename `let rec aux` to `let rec parseValuePathRest` or `let rec parseValuePathTail` to make it more clear inside `parseValuePath`? What do you think?

Issue #138, opened 2020-09-15
Differentiate between fn(.) and fn(.())


ryyppy:
Extracted from [this issue](https://github.com/facebook/reason/issues/2357):

>Currently they are all de-sugared into OCaml syntax fn () [@bs]
>
> The thing is that in BuckleScript, since there is no syntax space for arity zero, so we make a special form that fn () [@bs] is treated arity zero application.
>
> I wonder Reason could make a difference here which feels more intuitive:
> 
> fn (.) --> arity zero -> fn () [@bs]
> fn (. ()) --> arity 1 -> let unit = () in fn unit of course, some care need to be taken to avoid name collision

**Example where this causes an issue:**

```
Js.Promise.make((~resolve, ~reject as _) => Js.Global.setTimeout(() => {
      resolve(. ())
}, 1500)->ignore)
```

[See Playground Link](https://rescript-lang.org/try?code=DYUwLgBA7ghglpAvBAFASgogfAKAhAKQGcA6ABQCcB7AWziJBJpgGsQUUA-CkIq4AG4gANBG4gAViADGkGEQgB9DNkKkA4sCoAjGMBIMwAFTg0QVAK5gOKrBADeefM559B7EqjRon+AL6iAIwArAAMoWgAtFhwAOYAdlQ8PkA)

PR #139, opened 2020-09-15
implement syntax for arity zero vs arity one in uncurried application

IwanKaramazow:
Since there is no syntax space for arity zero vs arity one, we parse
  `fn(. ())` into
  `fn(. {let __res_unit = (); __res_unit})`
  when the parsetree is intended for type checking

`fn(.)` is treated as zero arity application

Fixes https://github.com/rescript-lang/syntax/issues/138

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+* Implement syntax for arity zero vs arity one in uncurried application in [#139](https://github.com/rescript-lang/syntax/pull/139)
 * Fix parsing of first class module exprs as part of binary/ternary expr in [#256](https://github.com/rescript-lang/syntax/pull/256)
 
 ## ReScript 9.0.0
--- a/src/res_ast_conversion.ml
+++ b/src/res_ast_conversion.ml
@@ -318,6 +318,11 @@ let stringLiteralMapper stringData =
     )
   }
 
+let hasUncurriedAttribute attrs = List.exists (fun attr -> match attr with
+  | ({Asttypes.txt = "bs"}, Parsetree.PStr []) -> true
+  | _ -> false
+) attrs
+
 let normalize =
   let open Ast_mapper in
   { default_mapper with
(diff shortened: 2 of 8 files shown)


ryyppy:
Not sure if there is any difference, but it might also be worth considering converting it into `fn(. ignore())`? At least it's easier to understand what's going on, and both versions compile to `fn(undefined)`

chenglou:
Hummmm not too sure

bobzhang:
@IwanKaramazow the proposal looks good to me.
`fn(.)` is compiled into `fn()`
`fn(.())` is `fn(undefined)`. For your short-cut implementation, it could be implemented as `fn(Pervasives.(())` -- this way you don't need introduce a special name

bobzhang:
would my proposal make the implementation a bit more clean? @IwanKaramazow

IwanKaramazow:
@bobzhang not 100%, it results in a type-error: https://rescript-lang.org/try?code=DYUwLgBAZgdhC8EAUA6CB9AlAgfMzA3AFBGyoDeADAUpgL6ZA

It could be changed to: `fn(. {ignore(0); ()})`

bobzhang:
@IwanKaramazow it is because we are more strict with sequence types. @ryyppy 's proposal also makes sense, the idea is not to introduce a magic name in the generated code

ryyppy:
@IwanKaramazow @bobzhang I don't understand how `fn(. {ignore(0); ()})` is cleaner than `fn(. ignore())`? Is there any benefit of having this local scope with two expressions?

I personally use `fn(. ignore())` and we also recommend it [in the docs](https://rescript-lang.org/docs/manual/latest/function#uncurried-function) right now, so ppl have a reference on how to work around the syntactical limitation.

IwanKaramazow:
Yea, let’s go with ignore(). Will finish this tomorrow

IwanKaramazow:
Ok, this PR is now ready to be merged.

The semantics are summarised by the following table:
image
Unfortunately, this is a breaking change. If you look at the table: `f(. ())` used to be parsed as `f(.)` (arity0). With this PR `f(. ())` would be parsed as `f(. ignore())` (arity1).
@bobzhang do we schedule this for the next breaking changes release?

bobzhang:
@IwanKaramazow I am happy to merge it, would you resolve the conflicts?

IwanKaramazow:
@bobzhang rebased! Do we merge?

bobzhang:
yes!

PR #146, opened 2020-09-20
Update parser + test runner to fail when in non-recover mode

jchavarri:
Extracted from #124.

- Makes parser cli return exit code 1 if parsed result is invalid (except when `-recover` is passed)
- Adapt runner to check status after calling `rescript.exe`, fail test if status > 0
- Update test `parsing/grammar/expressions/polyvariant.js ` that was failing to parse


IwanKaramazow:
@jchavarri perfect, nothing to add.

Issue #190, opened 2020-11-25
Formatting of function call with function with annotated labeled arguments passed as argument


alexfedoseev:
```rescript
// actual
let make = fn((~a: option<string>=?, ~b: option<string>=?, ~c: option<string>=?, ~d: option<
  string,
>=?, ~e: option<string>=?, x) => {
  ()
})

// expected: arguments bit
~a: option<string>=?, 
~b: option<string>=?, 
~c: option<string>=?, 
~d: option<string>=?, 
~e: option<string>=?, 
x
```

PR #199, opened 2020-12-03
Tweak printing of callbacks to break correctly in case of annotated labeled arguments

IwanKaramazow:
Fixes https://github.com/rescript-lang/syntax/issues/190

```rescript
// before
let make = fn((~a: option<string>=?, ~b: option<string>=?, ~c: option<string>=?, ~d: option<
  string,
>=?, ~e: option<string>=?, x) => {
  ()
})

// after
let make = fn((
  ~a: option<string>=?,
  ~b: option<string>=?,
  ~c: option<string>=?,
  ~d: option<string>=?,
  ~e: option<string>=?,
  x,
) => {
  Js.log()
})
```

### DIFF
--- a/src/res_doc.ml
+++ b/src/res_doc.ml
@@ -234,7 +234,7 @@ let debug t =
   let rec toDoc = function
     | Nil -> text "nil"
     | BreakParent -> text "breakparent"
-    | Text txt -> text ("text(" ^ txt ^ ")")
+    | Text txt -> text ("text(\"" ^ txt ^ "\")")
     | LineSuffix doc -> group(
         concat [
           text "linesuffix(";
@@ -245,6 +245,7 @@ let debug t =
           text ")"
         ]
       )
+    | Concat [] -> text "concat()"
     | Concat docs -> group(
         concat [
           text "concat(";
@@ -312,7 +313,7 @@ let debug t =
           indent (
             concat [
               line;
-              text ("shouldBreak: " ^ (string_of_bool shouldBreak));
+              text ("{shouldBreak: " ^ (string_of_bool shouldBreak) ^ "}");
               concat [text ",";  line];
               toDoc doc;
             ]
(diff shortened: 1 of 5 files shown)


IwanKaramazow:
@alexfedoseev what do you think of this layout to fix your issue?

IwanKaramazow:
I wonder if it would be better to have:
```rescript
let make = fn(
  (
    ~a: option<string>=?,
    ~b: option<string>=?,
    ~c: option<string>=?,
    ~d: option<string>=?,
    ~e: option<string>=?,
    x,
  ) => {
    Js.log()
  }
)
```

jfrolich:
Would be better IMO!

IwanKaramazow:
@jfrolich which one of the two are you referring to?

```rescript
// option 1
let make = fn((
  ~a: option<string>=?,
  ~b: option<string>=?,
  ~c: option<string>=?,
  ~d: option<string>=?,
  ~e: option<string>=?,
  x,
) => {
  Js.log()
})

// option 2
let make = fn(
  (
    ~a: option<string>=?,
    ~b: option<string>=?,
    ~c: option<string>=?,
    ~d: option<string>=?,
    ~e: option<string>=?,
    x,
  ) => {
    Js.log()
  }
)
```

jfrolich:
Option 2, makes it more clear that it's a function that receives a callback.

alexfedoseev:
Pro of 1: it doesn’t add +1 stop to a body.
Pro of 2: it’s clearer that argument is a function.

Personally, I’m fine with either. Most likely additional stop won’t be an issue in majority of cases (ppx’es are implemented in reason/ocaml anyway 😃)

IwanKaramazow:
Investigated and option 1 appears to be non-trivial. I'll merge to get the bug fixed and revisit this problem at a later date. Need to get infix operators done. Tracked further in: https://github.com/rescript-lang/syntax/issues/201

Issue #225, opened 2021-01-07
Numbers in escaped poly variant strings are reprinted without escape quotes


ryyppy:
Following snippets are currently dropping the `\""` sequences when formatted:

```rescript
let a = #\"1"
let b = #\"1a"
```

These snippets **don't remove** the escape sequence, although it's save to do:

```
// This escape could be removed
let c = #\"a2"

// Interestingly, this seems to remove the escape sequence correctly
let d = #\"abcd"
```

[Playground Link](https://rescript-lang.org/try?code=DYUwLgBAhhC8EGIA6AiAjCgUKSAjOiqaUW24EAxgcilAExY4QAm1qUuFzKQA)

### Expected behavior:

```rescript
let a = #\"1"
let b = #\"1a"

let c = #\"a2"
let d = #\"abcd"
```

should print to

```rescript
let a = #\"1"
let b = #\"1a"

let c = #a2
let d = #abcd
```

PR #228, opened 2021-01-08
Light weight exotic poly variant syntax

IwanKaramazow:
Implements https://github.com/BuckleScript/syntax/issues/76 and fixes https://github.com/rescript-lang/syntax/issues/225

The syntax for exotic identifiers in poly-variants is quite heavy: `#\"ease-in"`.
By extending the syntax for poly-variants with a string syntax, we get a lighter weight version that is more pleasant to read and write:
```
poly-var ::=
  HASH STRING
```
Example:
```ocaml
type animation = [ #"ease-in" | #"ease-out" ]
```

There was a previous [PR](https://github.com/rescript-lang/syntax/pull/96) for this. That one didn't ship as we weren't 100% sure of all the implications. Poly variants are essentially strings now, so this would make a great fit. Let's do this.

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -444,6 +444,164 @@ let processUnderscoreApplication args =
   in
   (args, wrap)
 
+let hexValue x =
+  match x with
+  | '0' .. '9' ->
+    (Char.code x) - 48
+  | 'A' .. 'Z' ->
+    (Char.code x) - 55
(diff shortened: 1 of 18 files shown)


ryyppy:
Tests are looking good from my side

bobzhang:
looks great

Issue #185, opened 2020-11-24
bs.open does not convert from ReasonML


jfrolich:
```reason
exception Error(string);

let handlePromiseFailure =
  [@bs.open]
  (
    fun
    | Error(_) => ()
  );
```

Formats into

```rescript
exception Error(string)

let handlePromiseFailure = x =>
  switch x {
  | Error(_) => ()
  }
```

The whole `@bs.open` is gone. I also have no idea how to express the code in the new rescript syntax.

https://rescript-lang.org/try?ext=re&code=KYDwxsAOAuCWD2A7ABAUQE7vugFAZ2nVkQHMBKAbgCgqAbYaZACwENEATegBSwFtY8wAGItYtAK7pgyALxVkyANoABAEZ4AdPEjBEAXXnIchhQDNxiE8gA+aTNhwB9MrIB8RsocpUgA

I think this is not a legacy way of handling exceptions as it has been introduced in 0.7 (https://github.com/rescript-lang/rescript-compiler/blob/master/site/docsource/Exception-handling.adoc#bsopen-type-safe-external-data-source-handling-since-170).

IwanKaramazow:
Sorry about this, we don't yet supported this in the new syntax.
You can leave it in the old syntax or write the code that `bs.open` expands to:
https://rescript-lang.org/try?code=KYDwxsAOAuCWD2A7ABAUQE7vugFAZ2nVkQHMBKAKAoBthpkALAQ0QBNaAFLAW1j2ABiTWNQCu6YMgC8yYJmzSAfBWTJYAM2QApPADpUIRLr4BhJt2oGIMBIgDy6O5GCIAakyItoOOVnRlkAG8VVWQ8AHdYaDAGZBw7ACMAK11uJhJYMB95fwAuWUMA4NDkAB80HOy-DiYAT2p4JlYAqUVkAGV4bmAq7Br6xuaQ1XKAfSVkADkkYGHkAF8Q+dlqfiC56cRZ1UWaOmQAN2ltPS4uvmBdCSTgMG9p6FH1eFE2Sj36cZkj0radXTOvH4ujATGiDF66CUIQiURijBY7GAgIuQhE4h6vmwRRC5U63RwIBaf1OPAuV2AeHg1AOPSJuKmMwm-xRwIkVJpPQARFzKDsKAFfmoSIhsLMgA

I'm currently focused on infix operators and some printer improvements, will look into this after those.

jfrolich:
Thanks!

[...]

PR #230, opened 2021-01-09
Print attributes without the @bs prefix

IwanKaramazow:
*before*
```rescript
@bs.val
@bs.module
```

*after*
```rescript
@val
@module
```

Exceptions:
- @bs.open
- @bs.pipe.send

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+* Print attributes/extension without bs prefix where possible in [#230](https://github.com/rescript-lang/syntax/pull/230)
 * Cleanup gentype attribute printing [fe05e1051aa94b16f6993ddc5ba9651f89e86907](https://github.com/rescript-lang/syntax/commit/fe05e1051aa94b16f6993ddc5ba9651f89e86907)
 * Implement light weight syntax for poly-variants [f84c5760b3f743f65e934195c87fc06bf88bff75](https://github.com/rescript-lang/syntax/commit/f84c5760b3f743f65e934195c87fc06bf88bff75)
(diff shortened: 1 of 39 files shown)


pckilgore:
This is generally very cool and good but could it be all-or nothing across annotations? Having a consistently inferior syntax is better than an inconsistent-occasionally nicer syntax (which is already a massive issue with existing *.re code that must share a codebase with any .res for the foreseeable future).

IwanKaramazow:
@pckilgore can you explain yourself more? I included all annotations, except for `@bs.open` and `@bs.send.pipe` because of the latter being deprecated and only there for backwards compatibility.

pckilgore:
Even a deprecated construct might be supported for years.  Unless there's a technical reason to not convert to `@send.pipe`, `@bs.send.pipe` is going to **look** like invalid syntax, or cause folks to think that something might be wrong with the parser because it *usually* rewrites those things. And to the extent discouraging usage of send.pipe is a goal, I would suggest the compiler would be a nice and explicit place to handle that, as it can provide context and explanation vs. inconsistent inferred behavior.  

`@bs.open` doesn't matter much to me if it's invalid syntax entirely (that is, `@open` is invalid).  I suppose I'm a bit curious why it was called out in that case?

ryyppy:
@pckilgore we will document the bs.send.pipe and bs.open as deprecated attributes in our syntax widget, so ppl will still have a chance to understand what those are for, or why they are printed differently.

The compiler will also most likely print a depreciation warning at some point? \cc @bobzhang

pckilgore:
I suppose I'm still curious why not treat them consistently and obviate any possible confusion (and also avoid the need to clutter the documentation with information that will be irrelevant to most).

IwanKaramazow:
It's ok here and won't cause a lot of confusion. The documentation will contain both versions any way.
The printed code will look a bit less cryptic in the same sense that when a user writes `let (x) = ((1) +(2))`, the formatter outputs `let x = 1 + 2`.

Regarding the `@bs.open`, there's still an open issue that needs to be tackled first https://github.com/rescript-lang/syntax/issues/185

PR #235, opened 2021-01-15
WIP: React JSX PPX v4

rickyvetter:
This is part of the proposal discussed in https://forum.rescript-lang.org/t/rfc-rescript-react/901

There are still some changes that need to be made:

- [ ] clean up JSX3/4 calling and make change to caller in compiler
- [ ] move random raises to Location.raise_errorf, add tracking markers for reanalyze
- [ ] clean up ReasonReact references
- [ ] more thorough test cases

### DIFF
--- a/src/reactjs_jsx_ppx_v4.ml
+++ b/src/reactjs_jsx_ppx_v4.ml
@@ -7,6 +7,7 @@
   bsconfig.json. Specifically, there's a field called `react-jsx` inside the
   field `reason`, which enables this ppx through some internal call in bsb
 *)
+
 open Ast_helper
 open Ast_mapper
 open Asttypes
@@ -890,6 +891,9 @@ let jsxMapper () =
       let (children, argsWithLabels) = extractChildren ~loc ~removeLastPositionUnit:true callArguments in
       let (key, argsWithLabels) = extractKey ~loc argsWithLabels in
       let componentNameExpr = constantString ~loc id in
+      let componentExpr = Exp.apply ~attrs ~loc
+        (Exp.ident ~loc {loc; txt = Ldot (Lident "ReactDOM", "stringToComponent")})
+        [(nolabel, componentNameExpr)] in
       let childrenExpr = transformChildrenIfListUpper ~loc ~mapper children in
       let recursivelyTransformedArgsForMake = argsWithLabels |> List.map (fun (label, expression) -> ({loc; txt = Lident (getLabel label)}, isOptional label, mapper.expr mapper expression)) in
       let (jsxCall, args) = (match childrenExpr with
@@ -928,9 +932,9 @@ let jsxMapper () =
       (Exp.apply
         ~loc
         ~attrs
-        (Exp.ident ~loc ~attrs:(depIgnore loc) {loc; txt = Ldot (Lident "ReactDOMRe", jsxCall ^ "Keyed")})
+        (Exp.ident ~loc ~attrs:(depIgnore loc) {loc; txt = Ldot (Lident "React", jsxCall ^ "Keyed")})
         ([
-          (nolabel, componentNameExpr);
+          (nolabel, componentExpr);
           (nolabel, props);
           (nolabel, key)
         ]))
@@ -938,9 +942,9 @@ let jsxMapper () =
       (Exp.apply
         ~loc
         ~attrs
-        (Exp.ident ~loc ~attrs:(depIgnore loc) {loc; txt = Ldot (Lident "ReactDOMRe", jsxCall)})
+        (Exp.ident ~loc ~attrs:(depIgnore loc) {loc; txt = Ldot (Lident "React", jsxCall)})
         ([
-          (nolabel, componentNameExpr);
+          (nolabel, componentExpr);
           (nolabel, props)
         ]))
     in
@@ -1452,6 +1456,7 @@ let jsxMapper () =
 let rewrite_implementation (code: Parsetree.structure) : Parsetree.structure =
   let mapper = jsxMapper () in
   mapper.structure mapper code
+
 let rewrite_signature (code : Parsetree.signature) : Parsetree.signature =
   let mapper = jsxMapper () in
   mapper.signature mapper code
--- a/src/reactjs_jsx_ppx_v4.ml
+++ b/src/reactjs_jsx_ppx_v4.ml
@@ -27,29 +27,32 @@ let optional str = Optional str
 
 
 let mkPropConst ident = Typ.constr {loc = Location.none; txt = ident} []
+let mkSingleVariant rf = Rtag ({loc = Location.none; txt = rf}, [], true, [])
+let mkPropVariant row_fields = Typ.variant (List.map mkSingleVariant row_fields) Closed None
+
 let allowedProps = [
 ("children", {jsString = None; typeString = mkPropConst (Ldot (Lident "React", "element"))});
 ("key", {jsString = None; typeString = mkPropConst (Lident "string")});
-("ref", {jsString = None; typeString = mkPropConst (Ldot (Lident "ReactDOMRe", "domRef"))});
-("ariaCurrent", {jsString = Some("aria-current"); typeString = mkPropConst (Lident "page|step|location|date|time|true|false")});
+("ref", {jsString = None; typeString = mkPropConst (Ldot (Lident "ReactDOM", "domRef"))});
+("ariaCurrent", {jsString = Some("aria-current"); typeString = mkPropVariant ["page"; "step"; "location"; "date"; "time"; "true"; "false"]});
 ("ariaDetails", {jsString = Some("aria-details"); typeString = mkPropConst (Lident "string")});
 ("ariaDisabled", {jsString = Some("aria-disabled"); typeString = mkPropConst (Lident "bool")});
 ("ariaHidden", {jsString = Some("aria-hidden"); typeString = mkPropConst (Lident "bool")});
-("ariaInvalid", {jsString = Some("aria-invalid"); typeString = mkPropConst (Lident "grammar|false|spelling|true")});
+("ariaInvalid", {jsString = Some("aria-invalid"); typeString = mkPropVariant ["grammar"; "false"; "spelling"; "true"]});
 ("ariaKeyshortcuts", {jsString = Some("aria-keyshortcuts"); typeString = mkPropConst (Lident "string")});
 ("ariaLabel", {jsString = Some("aria-label"); typeString = mkPropConst (Lident "string")});
 ("ariaRoledescription", {jsString = Some("aria-roledescription"); typeString = mkPropConst (Lident "string")});
-("ariaAutocomplete", {jsString = Some("aria-autocomplete"); typeString = mkPropConst (Lident "inline|list|both|none")});
-("ariaChecked", {jsString = Some("aria-checked"); typeString = mkPropConst (Lident "true|false|mixed")});
+("ariaAutocomplete", {jsString = Some("aria-autocomplete"); typeString = mkPropVariant ["inline"; "list"; "both"; "none"]});
+("ariaChecked", {jsString = Some("aria-checked"); typeString = mkPropVariant ["true"; "false" ;"mixed"]});
 ("ariaExpanded", {jsString = Some("aria-expanded"); typeString = mkPropConst (Lident "bool")});
-("ariaHaspopup", {jsString = Some("aria-haspopup"); typeString = mkPropConst (Lident "false|true|menu|listbox|tree|grid|dialog")});
+("ariaHaspopup", {jsString = Some("aria-haspopup"); typeString = mkPropVariant ["false"; "true"; "menu"; "listbox"; "tree"; "grid"; "dialog"]});
 ("ariaLevel", {jsString = Some("aria-level"); typeString = mkPropConst (Lident "int")});
 ("ariaModal", {jsString = Some("aria-modal"); typeString = mkPropConst (Lident "bool")});
 ("ariaMultiline", {jsString = Some("aria-multiline"); typeString = mkPropConst (Lident "bool")});
 ("ariaMultiselectable", {jsString = Some("aria-multiselectable"); typeString = mkPropConst (Lident "bool")});
-("ariaOrientation", {jsString = Some("aria-orientation"); typeString = mkPropConst (Lident "horizontal|vertical|undefined")});
+("ariaOrientation", {jsString = Some("aria-orientation"); typeString = mkPropVariant ["horizontal"; "vertical"; "undefined"]});
 ("ariaPlaceholder", {jsString = Some("aria-placeholder"); typeString = mkPropConst (Lident "string")});
-("ariaPressed", {jsString = Some("aria-pressed"); typeString = mkPropConst (Lident "true|false|mixed")});
+("ariaPressed", {jsString = Some("aria-pressed"); typeString = mkPropVariant ["true"; "false" ;"mixed"]});
 ("ariaReadonly", {jsString = Some("aria-readonly"); typeString = mkPropConst (Lident "bool")});
 ("ariaRequired", {jsString = Some("aria-required"); typeString = mkPropConst (Lident "bool")});
 ("ariaSelected", {jsString = Some("aria-selected"); typeString = mkPropConst (Lident "bool")});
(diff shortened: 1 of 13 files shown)


src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+let nolabel = Nolabel
+let labelled str = Labelled str
+let optional str = Optional str
+
+
+let mkPropConst ident = Typ.constr {loc = Location.none; txt = ident} []
+let mkSingleVariant rf = Rtag ({loc = Location.none; txt = rf}, [], true, [])
+let mkPropVariant row_fields = Typ.variant (List.map mkSingleVariant row_fields) Closed None
+
+let allowedProps = [
```
IwanKaramazow, line 23: Should we make this an array instead of a list?

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("content", {jsString = None; typeString = mkPropConst (Lident "string")});
+("controls", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("coords", {jsString = None; typeString = mkPropConst (Lident "string")});
+("data", {jsString = None; typeString = mkPropConst (Lident "string")});
+("dateTime", {jsString = None; typeString = mkPropConst (Lident "string")});
+("default", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("defer", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("disabled", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("download", {jsString = None; typeString = mkPropConst (Lident "string")});
+("encType", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 122: According to [HTML 5.2](https://www.w3.org/TR/html52/single-page.html#element-attrdef-form-enctype), this should be an enumerated attribute.
```suggestion
("encType", {jsString = None; typeString = mkPropVariant ["application/x-www-form-urlencoded"; "multipart/form-data"; "text/plain"]});
```

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("form", {jsString = None; typeString = mkPropConst (Lident "string")});
+("formAction", {jsString = None; typeString = mkPropConst (Lident "string")});
+("formTarget", {jsString = None; typeString = mkPropConst (Lident "string")});
+("formMethod", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 126: According to [HTML 5.2](https://www.w3.org/TR/html52/single-page.html#element-attrdef-submitbuttonelements-formmethod), this should be an enumerated attribute.
```suggestion
("formMethod", {jsString = None; typeString = mkPropVariant ["get"; "post"; "dialog"]});
```

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("headers", {jsString = None; typeString = mkPropConst (Lident "string")});
+("height", {jsString = None; typeString = mkPropConst (Lident "string")});
+("high", {jsString = None; typeString = mkPropConst (Lident "int")});
+("href", {jsString = None; typeString = mkPropConst (Lident "string")});
+("hrefLang", {jsString = None; typeString = mkPropConst (Lident "string")});
+("htmlFor", {jsString = None; typeString = mkPropConst (Lident "string")});
+("httpEquiv", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 133: According to [WHATWG HTML living standard](https://html.spec.whatwg.org/#attr-meta-http-equiv), this should be an enumerated attribute.
```suggestion
("httpEquiv", {jsString = None; typeString = mkPropVariant ["content-language"; "content-type"; "default-style"; "refresh"; "set-cookie"; "x-ua-compatible"; "content-security-policy"]});
```

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("icon", {jsString = None; typeString = mkPropConst (Lident "string")});
+("inputMode", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 135: According to [WHATWG HTML living standard](https://html.spec.whatwg.org/#input-modalities:-the-inputmode-attribute), this should be an enumerated attribute.
```suggestion
("inputMode", {jsString = None; typeString = mkPropVariant ["none"; "text"; "tel"; "url"; "email"; "numeric"; "decimal"; "search"]});
```

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("integrity", {jsString = None; typeString = mkPropConst (Lident "string")});
+("keyType", {jsString = None; typeString = mkPropConst (Lident "string")});
+("kind", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 138: According to [WHATWG HTML living standard](https://html.spec.whatwg.org/#attr-track-kind), this should be an enumerated attribute.
```suggestion
("kind", {jsString = None; typeString = mkPropVariant ["subtitles"; "captions"; "descriptions"; "chapters"; "metadata"]});
```

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("label", {jsString = None; typeString = mkPropConst (Lident "string")});
+("list", {jsString = None; typeString = mkPropConst (Lident "string")});
+("loop", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("low", {jsString = None; typeString = mkPropConst (Lident "int")});
+("manifest", {jsString = None; typeString = mkPropConst (Lident "string")});
+("max", {jsString = None; typeString = mkPropConst (Lident "string")});
+("maxLength", {jsString = None; typeString = mkPropConst (Lident "int")});
+("media", {jsString = None; typeString = mkPropConst (Lident "string")});
+("mediaGroup", {jsString = None; typeString = mkPropConst (Lident "string")});
+("method", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 148: According to [WHATWG HTML living standard](https://html.spec.whatwg.org/#attr-fs-method), this should be an enumerated attribute.
```suggestion
("method", {jsString = None; typeString = mkPropVariant ["get"; "post"; "dialog"]});
```

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("muted", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("name", {jsString = None; typeString = mkPropConst (Lident "string")});
+("nonce", {jsString = None; typeString = mkPropConst (Lident "string")});
+("noValidate", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("open_", {jsString = Some("open"); typeString = mkPropConst (Lident "bool")});
+("optimum", {jsString = None; typeString = mkPropConst (Lident "int")});
+("pattern", {jsString = None; typeString = mkPropConst (Lident "string")});
+("placeholder", {jsString = None; typeString = mkPropConst (Lident "string")});
+("poster", {jsString = None; typeString = mkPropConst (Lident "string")});
+("preload", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 161: According to [WHATWG HTML living standard](https://html.spec.whatwg.org/#attr-media-preload), this should be an enumerated variant.
```suggestion
("preload", {jsString = None; typeString = mkPropVariant ["none"; "metadata"; "auto"]});
```

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("radioGroup", {jsString = None; typeString = mkPropConst (Lident "string")});
+("readOnly", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("rel", {jsString = None; typeString = mkPropConst (Lident "string")});
+("required", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("reversed", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("rows", {jsString = None; typeString = mkPropConst (Lident "int")});
+("rowSpan", {jsString = None; typeString = mkPropConst (Lident "int")});
+("sandbox", {jsString = None; typeString = mkPropConst (Lident "string")});
+("scope", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 170: According to [WHATWG HTML living standard](https://html.spec.whatwg.org/#attr-th-scope), this should be an enumerated attribute.
I am not sure which is the best way to handle this, because the _auto state_ is not set by an explicit keyword, but the missing value (and invalid values, which it does not make sense in this context) is the _auto state_. Should we just use `mkPropVariant ["row"; "col"; "rowgroup"; "colgroup"]` or something different in order to allow being explicit about the _auto state_?

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("scoped", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("scrolling", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 172: According to [HTML 4.01](https://www.w3.org/TR/html401/present/frames.html#adef-scrolling), this can assume only fixed values. Note that this is an HTML 4 only attribute.
```suggestion
("scrolling", {jsString = None; typeString = mkPropVariant ["auto"; "yes"; "no"]});
```

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("srcSet", {jsString = None; typeString = mkPropConst (Lident "string")});
+("start", {jsString = None; typeString = mkPropConst (Lident "int")});
+("step", {jsString = None; typeString = mkPropConst (Lident "float")});
+("summary", {jsString = None; typeString = mkPropConst (Lident "string")});
+("target", {jsString = None; typeString = mkPropConst (Lident "string")});
+("type_", {jsString = Some("type"); typeString = mkPropConst (Lident "string")});
+("useMap", {jsString = None; typeString = mkPropConst (Lident "string")});
+("value", {jsString = None; typeString = mkPropConst (Lident "string")});
+("width", {jsString = None; typeString = mkPropConst (Lident "string")});
+("wrap", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 190: According to [WHATWG HTML living standard](https://html.spec.whatwg.org/#attr-textarea-wrap), this should be an enumerated attribute.
```suggestion
("wrap", {jsString = None; typeString = mkPropVariant ["soft"; "hard"]});
```

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("poster", {jsString = None; typeString = mkPropConst (Lident "string")});
+("preload", {jsString = None; typeString = mkPropConst (Lident "string")});
+("radioGroup", {jsString = None; typeString = mkPropConst (Lident "string")});
+("readOnly", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("rel", {jsString = None; typeString = mkPropConst (Lident "string")});
+("required", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("reversed", {jsString = None; typeString = mkPropConst (Lident "bool")});
+("rows", {jsString = None; typeString = mkPropConst (Lident "int")});
+("rowSpan", {jsString = None; typeString = mkPropConst (Lident "int")});
+("sandbox", {jsString = None; typeString = mkPropConst (Lident "string")});
```
dodomorandi, line 169: According to [WHATWG HTML living standard](https://html.spec.whatwg.org/#attr-iframe-sandbox), this should be
> an unordered set of unique space-separated tokens

which means that, in theory, we should take a _pseudo-list_ of a variant type that can be transformed into a string. I don't know if this can be expressed, and it's not the end of the world if it cannot be done -- just worth mentioning it.

dodomorandi (changes requested):
I tried to port some changed I proposed [in a PR for rescript-react](https://github.com/rescript-lang/rescript-react/pull/12). As I discussed in the original comment, this is just a _rough and simple_ check based on some comments left in the original `reason-react` code. Unfortunately, doing this specific check for all attributes is quite time consuming.

I added a reference for each attribute I commented, so you can easily check if I made any mistake, if you want.

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("ariaDetails", {jsString = Some("aria-details"); typeString = mkPropConst (Lident "string")});
+("ariaDisabled", {jsString = Some("aria-disabled"); typeString = mkPropConst (Lident "bool")});
+("ariaHidden", {jsString = Some("aria-hidden"); typeString = mkPropConst (Lident "bool")});
+("ariaInvalid", {jsString = Some("aria-invalid"); typeString = mkPropVariant ["grammar"; "false"; "spelling"; "true"]});
+("ariaKeyshortcuts", {jsString = Some("aria-keyshortcuts"); typeString = mkPropConst (Lident "string")});
+("ariaLabel", {jsString = Some("aria-label"); typeString = mkPropConst (Lident "string")});
+("ariaRoledescription", {jsString = Some("aria-roledescription"); typeString = mkPropConst (Lident "string")});
+("ariaAutocomplete", {jsString = Some("aria-autocomplete"); typeString = mkPropVariant ["inline"; "list"; "both"; "none"]});
+("ariaChecked", {jsString = Some("aria-checked"); typeString = mkPropVariant ["true"; "false" ;"mixed"]});
+("ariaExpanded", {jsString = Some("aria-expanded"); typeString = mkPropConst (Lident "bool")});
```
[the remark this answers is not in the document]
rickyvetter: Yeah I think these 3-state cases are where most issues will be here. When reading the spec I don't see a place where "undefined" has a behavior different from not any value at all. All true/false/undefined use undefined as default - so why provide an additional way to express when you get clean bools otherwise? I think if there are cases where the difference between not providing and providing an explicit `"undefined"` matters then it's a good change.

src/reactjs_jsx_ppx_v4.ml
```diff
@@ -0,0 +1,1455 @@
...
+("ariaHaspopup", {jsString = Some("aria-haspopup"); typeString = mkPropVariant ["false"; "true"; "menu"; "listbox"; "tree"; "grid"; "dialog"]});
+("ariaLevel", {jsString = Some("aria-level"); typeString = mkPropConst (Lident "int")});
+("ariaModal", {jsString = Some("aria-modal"); typeString = mkPropConst (Lident "bool")});
+("ariaMultiline", {jsString = Some("aria-multiline"); typeString = mkPropConst (Lident "bool")});
+("ariaMultiselectable", {jsString = Some("aria-multiselectable"); typeString = mkPropConst (Lident "bool")});
+("ariaOrientation", {jsString = Some("aria-orientation"); typeString = mkPropVariant ["horizontal"; "vertical"; "undefined"]});
+("ariaPlaceholder", {jsString = Some("aria-placeholder"); typeString = mkPropConst (Lident "string")});
+("ariaPressed", {jsString = Some("aria-pressed"); typeString = mkPropVariant ["true"; "false" ;"mixed"]});
```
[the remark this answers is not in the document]
rickyvetter: Similar as the other - does explicit "undefined" have different behavior from not providing ariaPressed at all? Much smaller downside to including here though.

src/reactjs_jsx_ppx_v4.ml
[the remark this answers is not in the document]
dodomorandi: What you say is totally right. The only hypothetical case (but it is *really* hypothetical) is when a function receive an element and spits it out after modifying it. If `ariaExpanded` was set, are we able to set it to `None`? If this is possible, even this very niche case can be solved, therefore the `undefined` variant can be avoided without consequences.

src/reactjs_jsx_ppx_v4.ml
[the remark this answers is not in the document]
dodomorandi: Yes, in this case we already *lost* the `bool` because of the `mixed` variant. If we don't need to express the `undefined` variant as discussed before, it is just a matter of choice. ☺️

cristianoc:
Continued here, in spirit: https://github.com/rescript-lang/syntax/pull/547

PR #248, opened 2021-01-24
Move towards "if" keyword for guards on pattern match cases.

IwanKaramazow:
### Motivation

Pattern match cases can accept guards to further refine the criteria for matching a case. Pattern guards appear after the pattern and consist of an expression following the `when` keyword. The grammar expresses this as `| PATTERN => EXPRESSION when EXPRESSION`. An example:
```rescript
switch person1 {
| Teacher(_) => () // do nothing
| Student({reportCard: {gpa}}) when gpa < 0.5 =>
  Js.log("What's happening")
| Student(_) =>
  // fall-through, catch-all case
  Js.log("Heyo")
}
```

We inherited the `when` keyword from our ancestors Reason and Ocaml. This keyword is not ideal:
1) It is not a keyword in JavaScript; you need to escape it in interop scenarios.
2) `when` takes up a keyword and is only being used in one small part of the grammar. (Cost/benefit)
3) Some people have trouble remembering the exact word, is it `when` or `where` etc.

Given the above, we're planning to move towards a new syntax for guards on match cases by using the `if` keyword.
```
match-guard ::=
  IF expr
```

Applied to the example above, this would give:
```diff
- | Student({reportCard: {gpa}}) when gpa < 0.5 =>
+ | Student({reportCard: {gpa}}) if gpa < 0.5 =>
```

### Migration path
1) Accept both `when` and `if` for case guards. Printer will output `if`. No breaking change.
2) Remove support for parsing case guards with the `when` keyword. Show a nice error when the user tries to use `when` instead of `if`

This PR implements `1)`

### Prior art

* [rust](https://doc.rust-lang.org/reference/expressions/match-expr.html#match-guards)

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -3230,7 +3230,7 @@ and parseWhileExpression p =
 
 and parsePatternGuard p =
   match p.Parser.token with
-    | When ->
+    | When | If ->
       Parser.next p;
       Some (parseExpr ~context:WhenExpr p)
     | _ ->
@@ -6132,7 +6132,7 @@ and parsePayload p =
       Parser.next p;
       let pattern = parsePattern p in
       let expr = match p.token with
-      | When ->
+      | When | If ->
         Parser.next p;
         Some (parseExpr p)
       | _ ->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+* Implement new syntax for guards on pattern match cases in [3248](https://github.com/rescript-lang/syntax/pull/248)
 * Implement intelligent breaking for poly-var type expressions in [#246](https://github.com/rescript-lang/syntax/pull/246)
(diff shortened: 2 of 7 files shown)


bobzhang:
reviewed, the diff is simple enough that lgtm

src/res_core.ml
```diff
@@ -3230,7 +3230,7 @@ and parseWhileExpression p =
 
 and parsePatternGuard p =
   match p.Parser.token with
-    | When ->
+    | When | If ->
```
bobzhang, line 3233: some nitpics: in general annotate the type p makes code more readable compared with `p.Parser.token`

IwanKaramazow: Thanks for the advice, will move towards the annotated style.

amiralies:
> 1. Accept both when and if for case guards. Printer will output if. No breaking change.


@IwanKaramazow I've to mention it's a breaking change since it breaks ci formatting checks.

IwanKaramazow:
@amiralies yes, good point. Has been released in a new major version: 9.0

PR #267, opened 2021-02-21
Fix outcome printer for locally abstract types

amiralies:


amiralies:
before
```rescript
'ae<'a> => 'a
```

after
```rescript
'a. e<'a> => 'a
```

IwanKaramazow:
@amiralies this is great! Thank you.

Would it be possible to add a test? You have to add an example cas in https://github.com/rescript-lang/syntax/blob/master/tests/oprint/oprint.res Make sure it type checkes (outcome printer runs on type checked trees), and the run `./lib/test.exe -update-snapshot`

jfrolich:
I also ran into this, great to see it fixed!

amiralies:
@IwanKaramazow  looks like the test is printing outcome signature which does not include local type information.
for example this snippet
```rescript
type rec poly<'a>

let f: type a. poly<a> => int = _ => 2
```
is being printed like this
```rescript
type rec poly<'a>
let f: poly<'a> => int
```
in the expected output

IwanKaramazow:
Ok, I'll merge 👍  Thanks again for the fix. Appreciate the effort.

IwanKaramazow:
(Will take care of adding your fix with your name in the changelog)

PR #268, opened 2021-02-22
Clean up scanner

chenglou:
- Try pattern matching on character in the scanner instead of using if-else with number constants
- Make scanner use char directly instead of converting back and forth between char and int

### DIFF
--- a/src/res_scanner.ml
+++ b/src/res_scanner.ml
@@ -12,7 +12,7 @@ type t = {
     -> endPos: Lexing.position
     -> Diagnostics.category
     -> unit;
-  mutable ch: int; (* current character *)
+  mutable ch: Char.t; (* current character *)
   mutable offset: int; (* character offset *)
   mutable rdOffset: int; (* reading offset (position after current character) *)
   mutable lineOffset: int; (* current line offset *)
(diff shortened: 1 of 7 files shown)


chenglou:
No alloc difference, maybe a tiny perf increase. Mostly only readability.

Before:
Benchmark •benchmarksdataRedBlackTree res rescript parser

After:
make bench

src/res_scanner.ml
```diff
@@ -54,28 +54,28 @@ let position scanner = Lexing.{
...
+  if scanner.rdOffset < Bytes.length scanner.src then (
     scanner.offset <- scanner.rdOffset;
     let ch = (Bytes.get [@doesNotRaise]) scanner.src scanner.rdOffset in
     scanner.rdOffset <- scanner.rdOffset + 1;
-    scanner.ch <- int_of_char ch
+    scanner.ch <- ch
   ) else (
     scanner.offset <- Bytes.length scanner.src;
-    scanner.ch <- -1
+    scanner.ch <- CharacterCodes.eof
```
chenglou, line 64: I've tried removing this hacky encoding by using a `isEndOfFile` field. But too many scanner helpers in this file rely on implicitly matching on some characters and sending the end-of-file `ch` case to an else branch. So using `isEndOfFile` is causing some infinite loops. Reverted back to hacking -1 into char.

We're not supposed to do this, but this isn't worst than what it is already.

PR #269, opened 2021-02-22
Remove some usages of bytes

chenglou:
Profiler showed that bytes allocation costed a bit. We don't mutate the src, so it can be a string. Fewer conversions

Benchmark of parser now faster! See screenshots of PR

### DIFF
--- a/src/res_comment.ml
+++ b/src/res_comment.ml
@@ -57,17 +57,16 @@ let trimSpaces s =
   let len = String.length s in
   if len = 0 then s
   else if String.unsafe_get s 0 = ' ' || String.unsafe_get s (len - 1) = ' ' then (
-    let b = Bytes.of_string s in
     let i = ref 0 in
-    while !i < len && (Bytes.unsafe_get b !i) = ' ' do
+    while !i < len && (String.unsafe_get s !i) = ' ' do
       incr i
     done;
     let j = ref (len - 1) in
-    while !j >= !i && (Bytes.unsafe_get b !j) = ' ' do
+    while !j >= !i && (String.unsafe_get s !j) = ' ' do
       decr j
     done;
     if !j >= !i then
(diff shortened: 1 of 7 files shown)


chenglou:
Before:
Screen Shot 2021-02-22 at 9 54 55 AM

After:
Screen Shot 2021-02-22 at 9 55 12 AM

src/res_comment.ml
```diff
@@ -57,17 +57,16 @@ let trimSpaces s =
   let len = String.length s in
   if len = 0 then s
   else if String.unsafe_get s 0 = ' ' || String.unsafe_get s (len - 1) = ' ' then (
-    let b = Bytes.of_string s in
     let i = ref 0 in
-    while !i < len && (Bytes.unsafe_get b !i) = ' ' do
+    while !i < len && (String.unsafe_get s !i) = ' ' do
```
chenglou, line 61: @IwanKaramazow why did you convert the string to byte here? You're only doing reads

src/res_io.ml
```diff
@@ -1,21 +1,15 @@
...
-      loop ()
-    )
+  let content =
+    try really_input_string chan (in_channel_length chan)
+    with End_of_file -> ""
   in
-  loop ()
+  close_in_noerr chan;
+  content
+
```
chenglou, line 9: So clean now

src/res_scanner.ml
```diff
@@ -118,7 +118,7 @@ let scanIdentifier scanner =
   ) do
     next scanner
   done;
-  let str = Bytes.sub_string scanner.src startOff (scanner.offset - startOff) in
+  let str = (String.sub [@doesNotRaise]) scanner.src startOff (scanner.offset - startOff) in
```
chenglou, line 121: `Bytes.sub_string` also raised exception, but reanalyze doesn't know yet. Fixed in reanalyze master

PR #272, opened 2021-02-24
CLEAN UP Scanner.scan EVEN MORE

chenglou:
The `scan` code was rather irregular, due to some quirks around when to `peek`, when to `next`, etc. I'm guessing out of a misguided desire to save perf.

The realization is that peeking 1 or 2 chars have the same perf, likewise for `next`ing 1 or 2 chars (well, in this diff, `next`ing 2 or 3 times are slightly more expensive but that's trivially fixable). Once we see that, we can just freely throw around the right amount of `peek` and `next` without needing to share some `next` code paths with some character lookahead and not with others.

Perf is actually better now! Screenshot in PR.

Also, much more readable. Previously, every time I check the `scan` code I needed to wrangle about which char I’m looking at and which char I’m looking at next, because of various imperative `next` calls here and there. No more!

### DIFF
--- a/src/res_scanner.ml
+++ b/src/res_scanner.ml
@@ -60,6 +60,29 @@ let position scanner = Lexing.{
   pos_cnum = scanner.offset;
 }
 
+(* Small debugging util
+❯ ./lib/rescript.exe baah/test.res
+let a = 1
+^ let
+let a = 1
+    ^ a
+let a = 1
+      ^ =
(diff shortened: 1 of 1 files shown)


chenglou:
Before:
Screen Shot 2021-02-24 at 4 30 29 AM

After:
Screen Shot 2021-02-24 at 4 30 44 AM

Issue #163, opened 2020-10-31
Leading single quote in constant expression is parsed silently


amiralies:
```rescript
let a = '"foo"
let b = 'true
let c = '2
let d = '2.2
let e = '''
````
is parsed into:
```rescript
let a = "foo"
let b = true
let c = 2
let d = 2.2
let e = '\''

```

amiralies:
I saw that there's a `SingleQuote` token in scanner. can we remove that and parse to more complex tokens directly?

IwanKaramazow:
Thanks for the report, will see what's best.

amiralies:
looks like it's a bit more general than what explained above.
for example this snippet
```rescript
let a =    =  =  = =  = =   '  ' }  } } } ] | | & * )   } =  = = =(x==y)
```
is being parsed into
```rescript
let a = x == y
```

amiralies:
Fixed via #279

PR #279, opened 2021-02-25
Fix atomic expressions parsing

amiralies:


amiralies:
Closes #163

IwanKaramazow:
Wow @amiralies, this is great! Thank you for taking the time!

PR #281, opened 2021-02-25
Scanner: clean up newline logic

chenglou:
This way things are less error prone
However... Having to check for newline every `next` regressed the perf back to the level after #269 😭... it's ok, we've earned it. There's been nothing but readability cleanup along the way so it's still a net win for reading.


chenglou:
So yeah, not sure

IwanKaramazow:
I kinda like the perf improvement =D

chenglou:
Alright I guess we'll just find a way to get it back somehow later

PR #283, opened 2021-02-26
Further code cleanup

chenglou:
It's unclear whether this is more readable, but whatever

### DIFF
--- a/src/res_scanner.ml
+++ b/src/res_scanner.ml
@@ -133,23 +133,27 @@ let make ~filename src =
     mode = [];
   }
 
+let isWhitespace ch =
+  match ch with
+  | ' ' | '\t' | '\n' | '\r' -> true
+  | _ -> false
+
 let rec skipWhitespace scanner =
-  match scanner.ch with
-  | ' ' | '\t'| '\n' | '\r' ->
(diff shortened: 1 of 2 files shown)


chenglou:
Before:
Screen Shot 2021-02-25 at 9 54 29 PM

After:

src/res_scanner.ml
```diff
@@ -133,23 +133,27 @@ let make ~filename src =
...
   let startOff = scanner.offset in
-  while (
-    CharacterCodes.isLetter scanner.ch ||
-    CharacterCodes.isDigit scanner.ch ||
-    '_' == scanner.ch ||
-    '\'' == scanner.ch
-  ) do
-    next scanner
-  done;
+  let rec skipGoodChars scanner =
```
IwanKaramazow, line 149: better assembly?

IwanKaramazow (approved):

src/res_scanner.ml
IwanKaramazow: Vs while loop

chenglou: No and I'm not optimizing at this level anymore, but you'll see

PR #285, opened 2021-02-26
Make Printer.classifyIdentContent much more readable

chenglou:
Not much perf difference. Slightly faster?


chenglou:
Before:
Screen Shot 2021-02-26 at 1 31 36 AM

After:
Screen Shot 2021-02-26 at 1 31 49 AM

Not much difference

IwanKaramazow (approved):

Issue #278, opened 2021-02-25
support object type spreading


bobzhang:
a.ml
```
type a = < x : int > 
type u = < a ; y : int >
```
npx bsc -format a.ml
```
type a = {"x": int}
type u = {, "y": int}
```

IwanKaramazow:
@bobzhang is there a use case for `type u = < a >`?

bobzhang:
No such use case.
Some corner case for testing:
```ocaml
type hi = < z : int >
type 'a u = < hi ; x : int ; y : int; .. >  as 'a
type 'a u1 = < hi; .. > as 'a
```

PR #290, opened 2021-02-27
Add support for object type spreading

IwanKaramazow:
Fixes https://github.com/rescript-lang/syntax/issues/278

```rescript
type a = {"x": int}
type u = {...a, "y": int}
```

### DIFF
--- a/tests/parsing/errors/other/__snapshots__/parse.spec.js.snap
+++ b/tests/parsing/errors/other/__snapshots__/parse.spec.js.snap
@@ -166,6 +166,9 @@ let record = { x with y }
 let { x; y } = myRecord
 let myList = x :: y
 let x::y = myList
+type nonrec t = < a  > 
+type nonrec t =
+  | Foo of < a  >  
 =====Errors=============================================
 
   Syntax error!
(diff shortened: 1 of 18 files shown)


bobzhang:
some test cases:
```res 
type hi = { "z" : int }
type  u <'a> = { ... hi ,  "x" : int , "y" : int,  .. }  as 'a
type  u1<'a> = { ... hi, ...hi, "y" : int , ...hi, ..  } as 'a
```

chenglou:
So I guess we said we'll disable `type a = {...b}` right?

tests/parsing/errors/other/__snapshots__/parse.spec.js.snap
```diff
@@ -216,14 +219,36 @@ Solution: you need to pull out each field you want explicitly.
...
   
   List pattern matches only supports one \`...\` spread, at the end.
 Explanation: a list spread at the tail is efficient, but a spread in the middle would create new list[s]; out of performance concern, our pattern matching currently guarantees to never create new intermediate data.
 
 
+  Syntax error!
+  parsing/errors/other/spread.js:10:11-13  
+   8 │ let list{...x, ...y} = myList
+   9 │ 
+  10 │ type t = {...a}
```
bloodyowl, line 236: Could the error be

```
You are using the spread `...` operator without extending the type, you can alias the type if you want it to be the same. 
```

or am I missing some case there?

IwanKaramazow: Thanks for the suggestion! Will implement.

IwanKaramazow: The general difficulty here is that we don't know whether it's an object or record…

bloodyowl: Yeah, in case the user makes the error on records, it'd give a two-step error:

```rescript
type b = {...a}
```

> **Error**: You are using the spread `...` operator without extending the type, you can alias the type if you want it to be the same.

```rescript
type b = {...a, x: string}
```

> **Error**: A record type declaration doesn't support the ... spread. Only an object (with quoted field names) does.

For sure it's a bit more cumbersome but I feel like this "teaches" the context where `...` applies better.

PR #291, opened 2021-02-27
Clean up Js.t object parsing, printing and converting

### LINKED ISSUE #263: Fix conversion of Js.t containing an alias core type
```reason
[@bs.deriving {abstract: light}]
type jsProps('a) = {
  config: Js.nullable(Js.t({..} as 'a)),
  image: Js.nullable(string),
  label: Js.nullable(string),
  maxwidth: Js.nullable(int),
};
```

The `Js.t({..} as 'a)` part is problematic. There's currently no way to express this in ReScript.
`{..}` is the syntax for `Js.t({..})`. So technically `{..} as 'a` (rescript), represents `Js.t({..)) as 'a`.
However in the above code the alias sits inside the `Js.t`.

### LINKED ISSUE #277: clean up object types handling
Currently `{"x" : int }` is parsed as `<x : int > Js.t`, 
on the master Js.t is a nop ` type 'a t = 'a`.
The parser can just parse it as ` < x : int >`

chenglou:
Fixes #277
Fixes #263

Now that https://github.com/rescript-lang/rescript-compiler/pull/4967 has landed:
- Parse `{"foo": int}` as ocaml `{. foo: int}`. Previously it parsed into ocaml `{. foo: int} Js.t`
- Remove a tiny printing optimizations for `Js.t`.
- For React's PPX 3, generate objects directly instead of `Js.t` objects. cc @rickyvetter @ryyppy for ppx4.
- The re->res converter automatically removes the `Js.t` part.
- Said converter has a bug (#263) that converts `Js.t({..}) as 'a` into `{..} as 'a` from naturally forgetting to special-case that path. Now this bug is conveniently ~~made into a feature~~ obsolete.

### DIFF
--- a/src/res_printer.ml
+++ b/src/res_printer.ml
@@ -1423,16 +1423,21 @@ and printTypExpr (typExpr : Parsetree.core_type) cmtTbl =
         doc
     in
     Doc.concat [typ; Doc.text " as "; Doc.concat [Doc.text "'"; printIdentLike alias]]
-  | Ptyp_constr({txt = Longident.Ldot(Longident.Lident "Js", "t")}, [{ptyp_desc = Ptyp_object (_fields, _openFlag)} as typ]) ->
-    let bsObject = printTypExpr typ cmtTbl in
-    begin match typExpr.ptyp_attributes with
-    | [] -> bsObject
-    | attrs ->
-      Doc.concat [
-        printAttributes ~inline:true attrs cmtTbl;
-        printTypExpr typ cmtTbl;
-      ]
-    end
+
+  (* object printings *)
+  | Ptyp_object (fields, openFlag) ->
+    printBsObjectSugar ~inline:false fields openFlag cmtTbl
+  | Ptyp_constr(longidentLoc, [{ptyp_desc = Ptyp_object (fields, openFlag)}]) ->
+    (* for foo<{"a": b}>, when the object is long and needs a line break, we
+       want the <{ and }> to stay hugged together *)
+    let constrName = printLidentPath longidentLoc cmtTbl in
+    Doc.concat([
+      constrName;
+      Doc.lessThan;
+      printBsObjectSugar ~inline:true fields openFlag cmtTbl;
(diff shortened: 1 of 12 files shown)


src/res_printer.ml
```diff
@@ -1423,16 +1423,21 @@ and printTypExpr (typExpr : Parsetree.core_type) cmtTbl =
...
-    | attrs ->
-      Doc.concat [
-        printAttributes ~inline:true attrs cmtTbl;
-        printTypExpr typ cmtTbl;
-      ]
-    end
+
+  (* object printings *)
+  | Ptyp_object (fields, openFlag) ->
+    printBsObjectSugar ~inline:false fields openFlag cmtTbl
```
IwanKaramazow, line 1429: Should we improve the name of this function?

tests/printer/typexpr/__snapshots__/render.spec.js.snap
```diff
@@ -327,7 +327,8 @@ external test: (
   foo,
   bar,
   baz,
-) => @attr {..
+) => @attr
```
IwanKaramazow, line 330: The line break is shady.

IwanKaramazow (approved):

IwanKaramazow:
Let's maybe also add a test for `Js.nullable(Js.t({..} as 'a))` conversion.

src/res_printer.ml
chenglou: Was thinking of unifying this with record printing, which we can discuss later. I'll do this now.

tests/printer/typexpr/__snapshots__/render.spec.js.snap
chenglou: I think before that, it went into a branch where it saw an attribute right before a type constructor (`Js.t`), so formatted that `Js.t` on the same line as the attribute, which then triggered the type constructor + object `<{` hugging branch.

But I've already discovered an existing error! `@foo {"bar": baz}` printing removes the `@foo`.

Issue #292, opened 2021-02-27
@bs.send doesn't reformat to @send


bloodyowl:
All the other bindings remove `bs.` on format, `@bs.send` remains the same. 

Seems to be the case with `@bs.inline` too

PR #293, opened 2021-02-28
Fix printing of @send and @inline attributes

IwanKaramazow:
Fixes https://github.com/rescript-lang/syntax/issues/292

Prints `@bs.send` as `@send` and `@bs.inline` as `@inline`

Followup of https://github.com/rescript-lang/syntax/pull/230


IwanKaramazow:
Superseded by https://github.com/rescript-lang/syntax/pull/294

PR #294, opened 2021-02-28
Clean up the conversion of @bs.* and %bs.*

### LINKED ISSUE #292: @bs.send doesn't reformat to @send

chenglou:
Fixes #292


chenglou:
Supersedes #293

IwanKaramazow (approved):

PR #304, opened 2021-03-03
Enable stack trace for binary; remove an `exit 1`

chenglou:
This makes the perf slightly slower (but basically unnoticeable), and the compilation slower, but this repo's files are compiled with stack trace enabled inside the compiler project anyway, so this is more realistic.

On the other hand, having a general try catch in res_cli that swallowed important exceptions such as `String.sub` and `Array.get` would have made debugging much harder.


chenglou:
Before:
Screen Shot 2021-03-03 at 7 06 25 AM

After:
Screen Shot 2021-03-03 at 7 06 37 AM

PR #306, opened 2021-03-05
Experimental: turn most of printer's lists into arrays

chenglou:
Excuse the commit messages. Just a toy.

This tries to allocate as few lists as possible.

Benchmark shows that everything allocates less (kinda obvious if you think about it), but the perf isn't necessarily better for now, mostly because the printer algorithm uses little pieces of lists (now arrays) again and again. So the saving of array isn't that big.

However, if/when we convert more things to array, like comments, maybe this'll shine more.


chenglou:
Before:
Screen Shot 2021-03-04 at 4 42 44 PM

After:
Screen Shot 2021-03-04 at 4 40 04 PM

PR #309, opened 2021-03-05
Printer test: optimize concat.

cristianoc:
Optimize concat to minimize allocations.
- join text docs early
- remove Nil
- flatten nested concats
- don't allocate unless the resulting list is not the same as the incoming one.


cristianoc:
On `Napkinscript.res rescript printer`:
before:
```
Allocs/op: 171960
```
after:
```
Allocs/op: 120221
```

chenglou:
Before:
image

After:
Screen Shot 2021-03-05 at 8 02 18 AM

chenglou:
Awesome work

PR #310, opened 2021-03-05
Printer experiment: version of `fits` that does not allocate.

cristianoc:


chenglou:
Before (reusing picture from last diff):


After:
Screen Shot 2021-03-05 at 8 36 32 AM

chenglou:
Slightly less readable but the benefit is huge enough for this to be merged. I'll comment it here and there. Thanks!

PR #313, opened 2021-03-06
Remove all old mentions of napkin

chenglou:
This was rescript's old name. No longer needed. Didn't touch the test files/cases mentioning it though.


tests/napkin_test.ml
```diff
@@ -1,207 +0,0 @@
-module IO = Res_io
```
chenglou, line 1: @IwanKaramazow this file is a duplicate of `res_test.ml` and isn't used?

IwanKaramazow: Remove if unused. Probably not converted

PR #316, opened 2021-03-08
Better outcome printer testing error message

chenglou:
Before, some errors during the snapshot test would be swallowed. That was confusing.

See PR for screenshots.


chenglou:
Before:
Screen Shot 2021-03-08 at 12 15 31 PM

After, when there's an unknown error:
image

With known error:
Screen Shot 2021-03-08 at 12 13 04 PM

PR #325, opened 2021-03-12
Printer: make propagateForcedBreaks allocation-free.

cristianoc:
Instead of returning a copy of the doc with some `IfBreak`s resolved, make doc mutable to mark when an `IfBreak` has been resolved to the first branch.

This makes the function allocation-free, but with side effects on the doc.

### DIFF
--- a/src/res_doc.ml
+++ b/src/res_doc.ml
@@ -15,10 +15,10 @@ type t =
   | Text of string
   | Concat of t list
   | Indent of t
-  | IfBreaks of {yes: t; no: t}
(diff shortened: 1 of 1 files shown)


chenglou:
Before:
Screen Shot 2021-03-15 at 5 41 39 PM
After:
Screen Shot 2021-03-15 at 5 42 57 PM

Damn...
Though making that doc mutable is slightly unfortunate. But the difference is huge!

IwanKaramazow (approved):
This makes actually more sense than the original immutable implementation.
The allocation gains are incredible again.

PR #327, opened 2021-03-15
Remove "to" and "downto" as keywords.

IwanKaramazow:
"to" is a commonly use identifier in interop scenarios. to/downto take a lot of expensive grammar real estate at a very high price. Since this is a handrolled parser, we can drop them as a keyword and parse them straight as identifier.


bobzhang:
note editor integration needs to be adapted cc @chenglou

Issue #303, opened 2021-03-03
jHi` causes String.sub exception


chenglou:
```res
jHi`
```

<img width="780" alt="image" src="https://user-images.githubusercontent.com/1909539/109785280-d573b780-7bc0-11eb-98b4-1cbc165c63db.png">

chenglou:
Same for:

```res
let a="
```

IwanKaramazow:
This is now fixed by https://github.com/rescript-lang/syntax/pull/328/files and https://github.com/rescript-lang/syntax/commit/c192534f9a08d81e66d54acf6da041a4af2eba80

PR #328, opened 2021-03-16
Fix & improve string error when quote is followed by eof

chenglou:
- Previously, `let a = "` followed by an eof would throw, despite `scanString` promising the opposite.
- This also includes the string's beginning quote into the error display (as opposed to only the content), so that such error doesn't highlight nothing.


src/res_scanner.ml
```diff
@@ -350,26 +350,32 @@ let scanStringEscapeSequence ~startPos scanner =
     ()
 
 let scanString scanner =
-  let offs = scanner.offset in
+  (* assumption: we've just matched a quote *)
```
chenglou, line 353: More dangerous here but it's ok. Only used in this file in one place.

src/res_scanner.ml
```diff
@@ -350,26 +350,32 @@ let scanStringEscapeSequence ~startPos scanner =
...
+  let firstCharOffset = scanner.offset in
 
-  let startPos = position scanner in
   let rec scan () =
     match scanner.ch with
-    | '"' -> next scanner
+    | '"' ->
+      let lastCharOffset = scanner.offset in
+      next scanner;
+      (String.sub [@doesNotRaise]) scanner.src firstCharOffset (lastCharOffset - firstCharOffset)
```
chenglou, line 364: No more `- 1` logic we need to double check!

tests/parsing/errors/scanner/__snapshots__/parse.spec.js.snap
```diff
@@ -165,11 +165,11 @@ exports[`unclosedComment.js 1`] = `
 
 exports[`unclosedString.js 1`] = `
 "=====Parsetree==========================================
-let z = \\"eof\\"
+let z = \\"eof\\\\n\\"
```
chenglou, line 168: This is a natural change but one can argue both ways; whether to include the newline into the highlight or not

tests/parsing/errors/scanner/__snapshots__/parse.spec.js.snap
```diff
@@ -165,11 +165,11 @@ exports[`unclosedComment.js 1`] = `
...
 =====Errors=============================================
 
   Syntax error!
-  parsing/errors/scanner/unclosedString.js:1:10-2:0  
+  parsing/errors/scanner/unclosedString.js:1:9-2:0
```
chenglou, line 172: @IwanKaramazow is there a case where we might need to highlight an ending `"`? I've only included the beginning `"`

src/res_scanner.ml
IwanKaramazow: Yes, this is an unofficial implicit assumption I regularly try to follow.

chenglou:
Fixes part of #303

tests/parsing/errors/scanner/__snapshots__/parse.spec.js.snap
IwanKaramazow: Hmm good question, my first reaction would be: unclosed strings will never have one. So we can't highlight it either way?

tests/parsing/errors/scanner/__snapshots__/parse.spec.js.snap
IwanKaramazow: How does this look in the terminal?

chenglou: Before:
Screen Shot 2021-03-16 at 12 22 00 AM

After:
Screen Shot 2021-03-16 at 12 22 38 AM

No visible difference visually for the existing highlighter

IwanKaramazow: The newline might be an artefact from my editor. Everything before the end-of-file token should be a part.

PR #333, opened 2021-03-18
[DoNotMerge] Test: propagate uses as stack an array that recycles slots.

cristianoc:
Implement the stack used by process as an extensible array where slots are recycled.
Adapt `fits` to also use the stack.


cristianoc:
This is just to remark that's it's possible to have more significant savings in the amount of allocation.
However, the changes are way too invasive, and it's unclear whether there is a speedup at all.
So this is just to illustrate a technique, though it's not effective in this case.

PR #343, opened 2021-03-25
Remove jest; simpler testing flow

chenglou:
Just a shell script sufficed

It basically just runs `rescript.exe` in parallel against all the test folders and output the result to some files in a nearby `expected`.
Then it uses git to check for changes and tells you about it. So that you can check git for changes =P

The question to ask is "what am I trying to do with snapshot tests?" and the answer for this repo is "I wanna check whether some files changed". After which you realize it's git's job.

Runtime went from 10s to 2s for regular tests, and 50-60s to 25s for roundtrip tests.

Snapshots are now just regular files, devoid of idiosyncrasies of snapshot tests' JS escaping.

I'll delete the remaining jest stuff in another diff.


chenglou:
Before:
Screen Shot 2021-03-24 at 6 24 53 PM

After:
Screen Shot 2021-03-24 at 6 25 47 PM

IwanKaramazow (approved):
Thanks again, the speed difference is a game changer. Let's do this.

PR #347, opened 2021-03-25
Update object method parsing and printer for ReScript 9

### LINKED ISSUE #311: object method call simplification
Thanks to the recent clean up, there is no need to apply a ppx for method call.

`obj["say"]` was currently compiled into `obj##say`, and transformed again in the compiler into Pexp_send with some wrappers.

This is not needed any more, it can be translated into `obj#say` without any post-processing.

To finish this, we should also implement the printer logic for `Pexp_send`

IwanKaramazow:
Fixes #311

`obj["say"]` was currently parsed as `obj##say` (Pexp_apply with ##)
We now parse this straight into `obj#say`(Pexp_send)

### DIFF
--- a/src/res_ast_conversion.ml
+++ b/src/res_ast_conversion.ml
@@ -421,7 +421,7 @@ let normalize =
                   pexp_attributes = [];
                   pexp_desc = Pexp_ident (Location.mknoloc (Longident.Lident "x"))
                 },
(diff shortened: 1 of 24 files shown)


IwanKaramazow:
Closed in favour of https://github.com/rescript-lang/syntax/pull/348

PR #349, opened 2021-03-25
Numeric poly variant support

IwanKaramazow:
Extends the grammar of poly variants with integers.

```rescript
let one = #1

switch numericPolyVar {
| #89 => ()
}

type permissions = [
  | #777
  | #644
]
```


bobzhang:
some more test cases: 

```
type t = [
  | #1 (string)
  | #2 (int, string)
]
``` 
Note this is not very useful but to be consistent && the backed treats them in a uniform way.

PR #350, opened 2021-03-26
Expose Res_outcome_printer.printOutTypeDoc / printOutSigItemDoc

ryyppy:
This particular APIs are needed to be able to print typedtree data
in the playground bundle. We need that to show type hint tooltips.


IwanKaramazow (approved):
LGTM

PR #387, opened 2021-04-14
not display external garbage

bobzhang:
```
jscomp$../darwin/bsc.exe test/debug_tmp.cmi 
external f: int => int = "xx" "����
��A@�"xx@@@"
jscomp$../darwin/bsc.exe test/debug_tmp.cmi 
external f: int => int = "xx" "#rescript-external"
```
Those characters are scaring

### DIFF
--- a/src/res_outcome_printer.ml
+++ b/src/res_outcome_printer.ml
@@ -598,7 +598,10 @@ let printPolyVarIdent txt =
                  Doc.text " =";
                  Doc.line;
                  Doc.group (
-                   Doc.join ~sep:Doc.line (List.map (fun prim -> Doc.text ("\"" ^ prim ^ "\"")) primitives)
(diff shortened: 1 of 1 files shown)


bobzhang:
cc @cristianoc 
the exception analysis seems could be improved:
```
  Exception Analysis
  File "./src/res_outcome_printer.ml", line 579, characters 11-29
  printOutSigItemDoc might raise Invalid_argument (res_outcome_printer.ml:602:51) and is not annotated with @raises Invalid_argument
```

cristianoc:
Improved how?

bobzhang:
We already did the check that it is not an empty string, so indexing the 0 character should not throw.
This is a false alarm

cristianoc:
That's not how an exception analysis works. It's like types. You don't expect to not check an option type just because you happen to know it's not None.

Issue #383, opened 2021-04-12
Explicit type annotation for unit


rickyvetter:
[`let foo = ((): unit)`](https://rescript-lang.org/try?code=DYUwLgBAZg9jEF4IApkEoBcECuA7AlmGkA)

Gets an error:
The code is not valid ReScript syntax.
[E] Line 1, column 19:
Did you forget a `=>` here?

bobzhang:
I am thinking of instead of fix it. How about giving the unit expression a new syntax instead of overloading `()`, for example `Void`.
Why I am proposing this is that `()` is already heavily loaded in tuple, call, and a bunch of other places, unit expression is rarely used, so maybe we can give it a new name instead

IwanKaramazow:
It would still be good to also fix the underlying problem. 
Interesting note, this is also a bug in TypeScript:
<img width="507" alt="image" src="https://user-images.githubusercontent.com/10634678/115113647-98565f80-9f8b-11eb-8fd5-d73acb97f78f.png">

It's a bit unfortunate that JS added the arrow function to their language. From a parser perspective the implementation needs to have a lookahead. Anyway, looking into the underlying issue.

a-c-sreedhar-reddy:
Does `()` in typescript mean void?

If not then the above error message in TS is not a bug.

phenax:
@IwanKaramazow, `let x = ()` is not valid typescript either so it isn't related.

Also, this seems to work fine -
```rescript
let foo = (():> unit)
```

I'm guessing [this](https://github.com/rescript-lang/syntax/blob/1680ae5d0e5546f5a150e028a148aeee257bcd20/src/res_core.ml#L2292) is the culprit.

IwanKaramazow:
@phenax uf yes, forgot that `()` doesn't exist in typescript. That syntax error makes total sense.

As @a-c-sreedhar-reddy diagnosed correctly in https://github.com/rescript-lang/syntax/pull/389, the parser thinks it's an arrow function.

@bobzhang the `()` is indeed super overloaded, good point. Maybe we could start thinking about an alternative syntax in the long term?

IwanKaramazow:
Fixed in https://github.com/rescript-lang/syntax/pull/389

Issue #359, opened 2021-04-01
can not convert such reason syntax into rescript syntax


bobzhang:
```reason
let fn = f => f;
type f = int => unit;
let a = fn(_ => (): f);

```

PR #389, opened 2021-04-15
Explicit type annotation for unit

a-c-sreedhar-reddy:
fixes - #383


IwanKaramazow (commented):
Hey @a-c-sreedhar-reddy, thanks for picking this up =D 

I wonder if we can solve this in a more general way. What if someone aliases `type u = unit` and then writes `let foo = ((): u)`? 

Maybe we can just see if the token after the `:` is an `Lident`, consume the token and then check if there's a `=>` ?

a-c-sreedhar-reddy:
Hi @IwanKaramazow . I have added the changes.  

If it is after 
 Colon, Lident _ if the next token is EqualGreater we return true else false.

a-c-sreedhar-reddy:
This would fail for this case 
```
type d<'a,'b>
let c = ():d<'a,'b> => 9
```
How can I parse the entire type expression after Colon and then search for => instead of using Lident?

a-c-sreedhar-reddy:
Hi @IwanKaramazow 
I have updated the PR.
instead of checking for `=>` immediately after `Lident`, we would skip `<int,int>` parts and then look for `=>`.

IwanKaramazow:
Thanks for the work @a-c-sreedhar-reddy, we're getting there. Next step would be to add some tests in `tests/parsing/grammar/expressions/arrow.res`.

We can also add the test case from https://github.com/rescript-lang/syntax/issues/359, which suffers from the same problem.

Could you also add a comment above every branch where you match on a token with an example?
Example:
```
match token with
(* arrived at `()` here *)
| Rparen
```

> How can I skip the entire type expression after the Colon and then search for => instead of using Lident?

The detection of arrow functions works based on heuristics, we could do a full parse. But I don't know if it's worth the additional complexity. Handling the most common cases seems like a good first step. How do you feel about this?

a-c-sreedhar-reddy:
> Next step would be to add some tests in `tests/parsing/grammar/expressions/arrow.res`

Hi @IwanKaramazow I have added the tests and also the comments.

IwanKaramazow:
Awesome thanks! Trying to think of additional edge cases, but I think this looks ready.

IwanKaramazow:
@a-c-sreedhar-reddy  Thanks for your work! Really appreciate the effort. Let's merge.

a-c-sreedhar-reddy:
Thanks @IwanKaramazow

PR #392, opened 2021-04-17
No need to cache npm installs anymore + use npm ci

chenglou:
We have a single dependency that's faster to install than to cache lol
Also, use npm ci, which is a subset of npm install that only reads into
the lockfile and doesn't do any user-facing adjustments.


IwanKaramazow:
image

PR #394, opened 2021-04-24
Fix regression in template literal string scanning.

IwanKaramazow:
The scanner underwent extensive refactoring, it appears that something fishy was introduced in d20c137d
Upon encountering a `$` character, the routine that scans template literal tokens, shouldn't consume two characters.

Example:
```rescript
`$${dollarAmountInt}`
```
Seeing a `$` doens't imply that the next character (here also a `$`) should be consumed.


JasoonS:
Amazing,  thanks Maxim!

PR #404, opened 2021-05-05
Fix printing of Osig_module in outcome printer.

IwanKaramazow:
Modules types don't use a `=`, it should be a `:`.

```
module Expr: {
  …
}
```


cristianoc:
Needs syncing to the editor right.

IwanKaramazow:
yes, should be synced with editor + compiler.

cristianoc:
Great. Done.

Issue #407, opened 2021-05-07
nulary polymrophic variants starting with leading number are compiled to numbers


amiralies:
for example
```rescript
let a = #10s
```
is compiled to 
```js
va a = 10
```

as result if this if I have a type containing `#10s` and `#10m`, these two values are eqivalent

IwanKaramazow:
Yes, confirmed the bug on my side, should be matching on `  | Int {i; suffix = None} ->`. Int like tokens with a suffix don't qualify as numeric polyvars.

Will make a PR.

PR #408, opened 2021-05-09
Don't parse Int token with suffices as hash ident for poly variants

IwanKaramazow:
`#10s` should not be accepted as a numeric polyvariant identifier.

Fixes https://github.com/rescript-lang/syntax/issues/407

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -124,6 +124,9 @@ Solution: directly use `concat`."
 
   let sameTypeSpread =
     "You're using a ... spread without extra fields. This is the same type."
+
+  let polyVarIntWithSuffix number =
+    "A numeric polymorphic variant cannot be followed by a letter. Did you mean `#" ^ number ^ "`?"
 end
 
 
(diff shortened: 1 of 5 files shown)


src/res_core.ml
```diff
@@ -77,7 +77,7 @@ Explanation: since records have a known, fixed shape, a spread like `{a, ...b}`
 Explanation: lists are singly-linked list, where a node contains a value and points to the next node. `list[a, ...bc]` efficiently creates a new item and links `bc` as its next nodes. `[...bc, a]` would be expensive, as it'd need to traverse `bc` and prepend each item to `a` one by one. We therefore disallow such syntax sugar.\n\
 Solution: directly use `concat`."
 
-  let variantIdent = "A polymorphic variant (e.g. #id) must start with an alphabetical letter."
+  let variantIdent = "A polymorphic variant (e.g. #id) must start with an alphabetical letter or is a number (e.g. #742)"
```
bloodyowl, line 80:
```suggestion
  let variantIdent = "A polymorphic variant (e.g. #id) must start with an alphabetical letter or be a number (e.g. #742)"
```

IwanKaramazow:
Yes, thanks for the suggestion, will do!

tests/parsing/errors/other/expected/hashIdent.res.txt
```diff
@@ -0,0 +1,37 @@
+
+  Syntax error!
+  tests/parsing/errors/other/hashIdent.res:1:10-12
+
+  1 │ let x = #10s
+  2 │ 
+  3 │ type t = [ #red | #10s ] 
+
+  A numeric polymorphic variant cannot be followed by a letter. Did you mean `#10`?
```
IwanKaramazow, line 9: @bloodyowl what do you think of this new message? Specialized it to give a clear suggestion on what the user could do in this case.

bloodyowl: Yup, that looks better 👍

Issue #412, opened 2021-05-14
Not allowing such syntax


bobzhang:
```
external f : (int,int)=>int = "hi" "hx"
```
This seems to not make much sense, hope to reduce some complexity budget

IwanKaramazow:
Yes, will remove. This would also drastically simplify that part of the grammar

PR #415, opened 2021-05-15
Tweak grammar of primitive list in externals

IwanKaramazow:
Fixes https://github.com/rescript-lang/syntax/issues/412

Currently the grammar allows for a list of primitives in an external declaration: i.e. `"hi" "hx"` in `external f: (int, int) => int = "hi" "hx"`. This stems from the fact that user primitives with arity greater than 5 should be implemented by two C functions. The first function, to be used in conjunction with the bytecode compiler ocamlc, receives two arguments: a pointer to an array of OCaml values (the values for the arguments), and an integer which is the number of arguments provided. The other function, to be used in conjunction with the native-code compiler ocamlopt, takes its arguments directly.

However in the case of compiling to JS with ReScript, we don't need to deal with this. In order to reduce some complexity, we'll now parse just one primitive.


IwanKaramazow:
cc @bobzhang

tests/parsing/errors/structure/expected/external.res.txt
```diff
@@ -0,0 +1,11 @@
+
+  Syntax error!
+  tests/parsing/errors/structure/external.res:2:51
+
+  1 │ // missing JS value name
+  2 │ external setTimeout: (unit => unit, int) => float =
+  3 │ 
+
+  An external requires the name of the JS value you're referring to, like "setTimeout".
```
IwanKaramazow, line 9: cc @bloodyowl, how does this sound? 😄

bloodyowl: Does this only occur with:

```
external setTimeout: (unit => unit, int) => float =
```

or also with

```
external setTimeout: (unit => unit, int) => float
```

?

IwanKaramazow: Currently only implemented for the first. Can also add one for the latter

bobzhang:
we may also need consider multiple line string support:
```
external add : (int,int)=>int = `function add (a,b){
   return a + b
}`
```

IwanKaramazow:
Yes, `"` supports multiline strings because of backwards compatibility. I can add support for ` strings, but there's currently no way to encode this in the ast. Makes it hard for the printer. Let's stick with `"` strings in the short term?

bobzhang:
sounds good, as long as we have a way to encode multiple line string, it should be fine

tests/parsing/errors/structure/expected/external.res.txt
bloodyowl: I'd go for:

```
2 | external setTimeout: (unit => unit, int) => float =
3 |
This external is missing the JavaScript value it is supposed to reference (like `= "setTimeout"`)
```

But the one is your PR looks good to me as well 😊

PR #416, opened 2021-05-15
Remove `{path as p}` record pattern sugar for `{path: p}`

IwanKaramazow:
`{path as p}` formats to `{path: p`}, which is the right syntax. The as syntax is unnecessarily confusing sugar.


IwanKaramazow:
cc @chenglou

PR #424, opened 2021-05-26
Better error message on capitalized type

ryyppy:
First attempt tackling #122 

There are a few tests breaking bc of this, not sure how to solve them.

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -728,8 +728,14 @@ let parseValuePath p =
   | Lident ident -> Longident.Lident ident
   | Uident ident ->
     Parser.next p;
-    Parser.expect Dot p;
-    aux p (Lident ident)
+    begin match p.token with
+      | Dot ->
+        Parser.next p;
+        aux p (Lident ident)
+      | _ ->
+        Parser.err p (Diagnostics.message "Identifier needs to be lower-cased.");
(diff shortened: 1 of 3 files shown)


src/res_core.ml
```diff
@@ -5223,7 +5229,20 @@ and parseTypeDefinitionOrExtension ~attrs p =
...
+  let name = match p.token with
+    | Uident ident ->
+      let identStartPos = p.startPos in
+      let identEndPos = p.endPos in
+      Parser.next p;
+      begin match p.token with
+        | Dot -> parseValuePathTail p startPos (Longident.Lident ident)
+        | _ ->
+          Parser.err p ~startPos:identStartPos ~endPos:identEndPos (Diagnostics.message "Type names need to start with a lower-cased letter.");
+          Location.mkloc (Longident.Lident ident) (mkLoc identStartPos identEndPos)
```
ryyppy, line 5242: Not sure if startPos / endPos is correct here

ryyppy:
Probably worth adding a suggestion in the error message as well: `Did you mean `foo`?` Should I try adding that?

IwanKaramazow:
Yes, definitely worth a try. In general I like having an error message follow the format "what" + "this-is-how-you-could-fix-it".

PR #425, opened 2021-05-26
Handle windows CRLF correct.

IwanKaramazow:
`\r\n` should be picked up as one line break, not two.


bobzhang:
tested on windows, looks good

ChangJoo-Park:
Thanks @IwanKaramazow

Issue #397, opened 2021-04-29
support for unicode characters


bobzhang:
let h = [':ship:', '世' , '界']

Assuming the source code is encoded in utf8, these characters should be parsed as Const_int (code_point).
I suggest we enforce that all source code must be encoded in utf8 which is similar to other languages like golang

jfrolich:
How would you handle the difference in encoding in the file (utf-8) and make it compatible with the representation in the JS engine (utf-16)? The above example would be a single utf-8 character but in the JS runtime it would be represented as two characters as emojis don't fit the 16 bits.

bobzhang:
there is a mis-conception. The source encoding is not relevant to string encoding, we choose utf8 just to follow the existing conventions, we do this which is exactly the same as ts/js: whose source are encoded in utf8 yet the string are utf16.
In rescript char will be int which is large enough for any code point.

IwanKaramazow:
@jfrolich if you're interested you can read more about it here: https://262.ecma-international.org/5.1/#sec-6

> If an actual source text is encoded in a form other than 16-bit code units it must be processed as if it was first converted to UTF-16.

The unicode topic is a bit confusing as there can be a difference between your javascript source file encoding and the encoding of strings at runtime. When you send JS files encoded as UTF-8 (which is for example the default when I open a JS file with VSCode) and send the appropriate headers, then the browser will decode the UTF-8.

@bobzhang will take care of this :+1:

jfrolich:
> there is a mis-conception. The source encoding is not relevant to string encoding, we choose utf8 just to follow the existing conventions, we do this which is exactly the same as ts/js: whose source are encoded in utf8 yet the string are utf16.
In rescript char will be int which is large enough for any code point.

I didn't say anything that contradicts what you are saying here, and I agree what you are saying. The javascript engine will transparently encode the file and external inputs from utf-8 to utf-16 (internal representation).

The the thing to reconcile is character size (a rescript char is 32-byte but a js char is always two byte). This also means that the emoji as stated above will be represented as two characters in JavaScript. This will cause issues when calculating the length of the string, or iterating over characters of a string. Perhaps @bobzhang has a solution already for this, but that was what I was alluding to.

IwanKaramazow:
Would it be possible to give a concrete example of what you're trying to do? Forgive me, but I always get infinitely confused about what length and character means in the context of encoded strings.

Let's say you have a plain JS file in UTF8 with an emoji.
[...]

PR #433, opened 2021-06-04
Unicode support

IwanKaramazow:
## Unicode support
This PR adds support for Unicode codepoints at the syntax level: ReScript source code is now considered unicode text encoded in UTF-8.

Fixes https://github.com/rescript-lang/syntax/issues/397

### Codepoint literals

A codepoint literal represents an integer value identifying a unicode code point. It is expressed as one or more characters enclosed in single quotes. Examples are `'x'`, `'\n'` or `'\u{00A9}'`. Multiple UTF-8-encoded bytes may represent a single integer value.

### String literals

String literals are (possibly multi-byte) UTF-8 encoded character sequences between double quotes, as in `"fox 🦊 \u{2665}"`. Internally they compile now as `{js|fox 🦊 \u{2665}|js}`, we don't want to garble the js output.

### New escape sequences

Both codepoint and string literals accept the following new escape sequences:

1) Unicode escape sequences
Any character with a character code lower than 65536 can be escaped using the hexadecimal value of its character code, prefixed with `\u`. Unicode escapes are six characters long. They require exactly four characters following `\u` . If the hexadecimal character code is only one, two or three characters long, you’ll need to pad it with leading zeroes.
Example: `'\u2665'` (Represents ♥)

2) Unicode codepoint escape sequences
Any code point or character can be escaped using the hexadecimal value of its character code, prefixed with `\u{` and suffixed with `}` . This allows for code points up to 0x10FFFF, which is the highest code point defined by Unicode. Unicode code point escapes consist of at least five characters. At least one hexadecimal character can be wrapped in `\u{…}` . There is no upper limit on the number of hex digits in use (for example `'\u{000000000061}' == 'a'`)
Example: `'\u{2318}'` (Represents ⌘)


jfrolich:
Why are `\u{..}` escape sequences needed in string literals if the source text is already assumed to be encoded in utf-8 (and thus able to express any code point)?

IwanKaramazow:
Escape sequences are primarily used to put hard/difficult to represent "characters" in a character or string literal. I guess you could compare it by using `\n`, `\t` or a hex escape sequence like `\x41`.

jfrolich:
Does this mean that normal string literals (`""`) when compiled to a JavaScript source file would be 100% compatible, as opposed to messing up the string literal when there is a codepoint that doesn't fit in 16-bit? (like an emoji)? That would be very good news!

IwanKaramazow:
@jfrolich Indeed that's the goal, no more weird escaping in the js output =D

jfrolich:
Question: why does this need to happen at the syntax level? Wouldn't it be a better fix to just don't let the rescript compiler fudge strings in any case and not just `{js|...|js}` type strings?

IwanKaramazow:
It happens on both levels:
- compiler: `{js|...|js}` ast node is a compiler construct with utf8 semantics. 
- syntax: needs to parse strings into that construct

Hongbo suggested to just parse strings as `{js|...|js}`, since it already contains all the logic.

jfrolich:
Yeah, I guess it works, but is there any reason to keep the garbled way of parsing string literals around in the compiler.

BTW: will it also print `{js|...|js}` as `"..."` with this change?

IwanKaramazow:
>  is there any reason to keep the garbled way of parsing string literals around in the compiler.

Time is limited, this is the fastest way to get the feature.

> the JS interpreter will automatically convert a UTF8 string literal to UTF16 internally, so if you want behavior that matches JS, the current approach doesn't make sense. Also if you want to match legacy OCaml behaviors the approach also doesn't make sense, because OCaml sees strings as bytes, so some kind of UTF16 internal representation doesn't make sense in that case as well.

The standard says that the runtime model of JavaScript is based on UTF-16 code units. That doesn't specify what byte-encoding is used to convert a source file to those units.

> BTW: will it also print {js|...|js} as "..." with this change?
The printer follows a different codepath with a separate internal encoding, you can rely on the parser emitting `{js|…|js}` for the compiler but not the other way around.

jfrolich:
> The standard says that the runtime model of JavaScript is based on UTF-16 code units. That doesn't specify what byte-encoding is used to convert a source file to those units.

Indeed. But most JavaScript files and thus string literals are encoded in utf-8 in the source files. If we need the garbling around (I never needed it and never found a usecase for it even after researching), can we put it as `{utf16|...|utf16}` for the people that really need it.

I am aware that this quick fix is probably the fastest way to get it shipped while not relying on Hongbo, but I am speaking on the long term. And if we agree I might be able to try and help accomplish this?

jfrolich:
> The printer follows a different codepath with a separate internal encoding, you can rely on the parser emitting {js|…|js} for the compiler but not the other way around.

Hmm so reformatting `"bla"` is going to yield `{js|bla|js}` (in some amount of the cases because it's not 100% reliable if I understand correctly?). That is not super pretty.

IwanKaramazow:
If you don't mind me asking, why do you you want your source files encoded in utf16?

UTF-8 is the most common character encoding method used on the internet today. [Over 97% of all websites](https://w3techs.com/technologies/cross/character_encoding/ranking), likely including your own, store characters this way.
UTF-8 encoding is preferable to UTF-16 on the majority of websites, because it uses less memory. Recall that UTF-8 encodes each ASCII character in just one byte. UTF-16 must encode these same characters in either two or four bytes. This means that an English text file encoded with UTF-16 would be at least double the size of the same file encoded with UTF-8.

> Hmm so reformatting "bla" is going to yield {js|bla|js} (in some amount of the cases because it's not 100% reliable if I understand correctly?). That is not super pretty.

`"bla"` will still yield `"bla"`.

jfrolich:
Haha no I don't want my source encoded in utf16 at all where did you think I said that? I am just saying that even if string literals are in utf-8 at the file level v8 (or any javascript engine) will convert them to the correct internal runtime representation (which is utf16 for JavaScript), so there is no need to do this "garbling" (let's use this term, probably not the best but you get what I mean).

I am just proposing the following (longer term):

- Don't run garbling on string at all at a compiler level for `""`.
- If we need to keep this around (I would love to see a usecase, so I know better why this feature exists, I think only Hongbo knows this), give it a separate literal like `{utf16|...|utf16}`.

jfrolich:
> `"bla"` will still yield `"bla"`.

Ok, that is nice!

Issue #432, opened 2021-06-03
Fatal error when using incorrect syntax with string interpolation


popstarfreas:
**Rescript version**: 9.1.3
Using certain incorrect syntax with string interpolation results in the compiler throwing an error with no helpful information.

**Code**
Example 1:
```rescript
`${a.}`
```

Example 2:
```rescript
`${{}`
```

Example 3:
```rescript
`${`
```

Compiler output:
`Fatal error: exception Invalid_argument("String.sub / Bytes.sub")`

bobzhang:
I will have a look

bobzhang:
after further look, it seems the interpolation syntax tries to solve a very difficult problem that I tried to avoid when designing the interpolation syntax in ml.

```res
`${ (a+b +c ) : int => int }`
```
it seems we allow arbitrary interoplaton syntax inside, which means itself could have ``` as well

popstarfreas:
I didn't see this repo when I posted this issue originally, but looking now, I just saw this existing issue: https://github.com/rescript-lang/syntax/issues/366
That issue looks to be the same as this one, and in 9.1.4, the compiler gives a better error for the example in that issue so I assume it is the same/related.

IwanKaramazow:
This is now fixed in https://github.com/rescript-lang/syntax/commit/c192534f9a08d81e66d54acf6da041a4af2eba80

PR #435, opened 2021-06-12
tentative fix for #432

bobzhang:


src/res_scanner.ml
```diff
@@ -487,11 +487,11 @@ let scanTemplateLiteralToken scanner =
       | _ ->
         next scanner;
         scan ())
-    | ch when ch == hackyEOFChar ->
+    | ch when ch = hackyEOFChar ->
       let endPos = position scanner in
       scanner.err ~startPos ~endPos Diagnostics.unclosedTemplate;
       Token.TemplateTail(
-        (String.sub [@doesNotRaise]) scanner.src startOff (scanner.offset - 1 - startOff)
+        (String.sub [@doesNotRaise]) scanner.src startOff (max (scanner.offset - 1 - startOff) 0)
```
bobzhang, line 494: for cases like
```
`${`
```
for the inner `, it is already the end of file, here both `startOff` and `offset` points to the end of the file, so the length was actually `-1`, let me know if you have a better fix

src/res_scanner.ml
```diff
@@ -487,11 +487,11 @@ let scanTemplateLiteralToken scanner =
       | _ ->
         next scanner;
         scan ())
-    | ch when ch == hackyEOFChar ->
+    | ch when ch = hackyEOFChar ->
```
IwanKaramazow, line 490: Why do you prefer `=` here?

IwanKaramazow (approved):
Thanks!

src/res_scanner.ml
bobzhang: `==` is not correct here, it happens to work since char is a immediate number, if it is float which is boxed in ocaml native this will be broken

PR #442, opened 2021-06-19
resdoc: add support for .ml and .res files

Minnozz:
Building on the [experimental-docgen](https://github.com/rescript-lang/syntax/tree/experimental-docgen) branch.

Fixes rescript-association/rescript-lang.org#359

### DIFF
--- a/.depend
+++ b/.depend
@@ -72,8 +72,8 @@ src/res_scanner.cmx : src/res_token.cmx src/res_diagnostics.cmx \
     src/res_comment.cmx src/res_character_codes.cmx src/res_scanner.cmi
 src/res_scanner.cmi : src/res_token.cmx src/res_diagnostics.cmi
 src/res_token.cmx : src/res_comment.cmx src/res_character_codes.cmx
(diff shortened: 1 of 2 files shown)


Minnozz:
A problem with the current `Ast_iterator` approach for structure files is that it recurses into functions, while we only want to document top level stuff.

I've been looking at a way to convert a `structure` into a `signature`, but I can't find it in the codebase. However, the new `dump` subcommand does something similar. Maybe we can use that?

IwanKaramazow:
* Regarding `Ast_iterator`, maybe it's ok to just write your own recursive functions that walk the tree? This way you have full control.

* Converting a `structure` into a `signature` is not possible without access to the type checker.

Minnozz:
@IwanKaramazow Thanks for the tips!

I initially thought using `Ast_iterator` would be better, but I think I'll revert to @ryyppy's approach.

IwanKaramazow:
(Random thought: Control flow is one of the only tools we have as programmers, sometimes it's not worth to give it up.)

PR #444, opened 2021-06-27
Fix outcome printing for empty objects

amiralies:


amiralies:
It should match parsing/printing of empty objects

IwanKaramazow:
Great! Thanks!!!

PR #454, opened 2021-08-23
Fix "depext unmet availability conditions" GH actions CI

IwanKaramazow:


IwanKaramazow:
Not sure why this suddenly broke… this seems to allow building and running tests on CI again…

Issue #472, opened 2021-12-21
Formatting a polymorphic variant with an empty string name causes a syntax error


tom-sherman:
Thank you for filing! Check list:
- [x] Is it a bug? Usage questions should often be asked in the [forum](https://forum.rescript-lang.org) instead.
- [x] Concise, focused, friendly issue title & description.
- [x] A [minimal, reproducible example](http://sscce.org).
- [x] OS and browser versions, if relevant.
- [x] Is it already fixed in master?

Steps to reproduce:

1. Format this code `let x = #""`

Expected behaviour:

Nothing or formatting applied

Actual behaviour:

`""` is stripped leaving you with `let x = #` which is a syntax error.

PR #474, opened 2021-12-30
fix: prevent stripping of quotes for empty polymorphic variant

a-c-sreedhar-reddy:
Fixes  https://github.com/rescript-lang/syntax/issues/472


IwanKaramazow:
Thanks!

PR #479, opened 2022-03-22
Fix parsing punned record fields with module name

cknitt:
This fixes https://github.com/rescript-lang/rescript-compiler/issues/5362.

I.e.

```rescript
{A.a, b}
```

is now correctly parsed to match the AST for

```rescript
{A.a: a, b: b}
```

and not for

```rescript
{A.a: A.a, b: b}
```

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -493,6 +493,12 @@ let hexValue ch =
   | 'A'..'F' -> (Char.code ch) + 32 - (Char.code 'a') + 10
   | _ -> 16 (* larger than any legal value *)
 
+(* Transform A.a into a. For use with punned record fields as in {A.a, b}. *)
+let removeModuleNameFromPunnedFieldValue exp =
+  match exp.Parsetree.pexp_desc with
+  | Pexp_ident pathIdent -> {exp with pexp_desc = Pexp_ident { pathIdent with txt = Lident (Longident.last pathIdent.txt) }} 
+  | _ -> exp
+
 let parseStringLiteral s =
   let len = String.length s in
   let b = Buffer.create (String.length s) in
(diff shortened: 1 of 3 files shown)


tests/parsing/grammar/expressions/expected/record.res.txt
```diff
@@ -2,8 +2,8 @@ let r = { a = expr }
 let r = { a = expr }
 let r = { Parsetree.pexp_attributes = [||]; Parsetree.loc = loc }
 let r = { a; b; c }
-let r = { Parsetree.pexp_attributes; Parsetree.loc }
-let r = { Parsetree.pexp_attributes; Parsetree.loc }
+let r = { Parsetree.pexp_attributes = pexp_attributes; Parsetree.loc = loc }
```
ryyppy, line 5: I think this shouldn't have changed?

cknitt: I think the AST is correct and that's just the way it is printed if "key != value expression".

IwanKaramazow: This change is indeed correct, the diff shows the correctly parsed AST.

src/res_core.ml
```diff
@@ -2711,6 +2711,11 @@ and parseJsxChildren p =
     (true, [parsePrimaryExpr ~operand:(parseAtomicExpr p) ~noCall:true p])
   | _ -> (false, loop p [])
 
+and removeModuleNameFromPunnedFieldValue exp =
```
IwanKaramazow, line 2714: Can you move this helper somewhere at the beginning of the file?
(Context, the mutually recursive functions are mostly concerned with actual parsing logic, not really transformations) 
Would maybe also be nice to have a comment above with what the transformation actually looks like.

src/res_core.ml
```diff
@@ -2946,7 +2951,7 @@ and parseRecordRow p =
       let fieldExpr = parseExpr p in
       Some (field, fieldExpr)
     | _ ->
-      Some (field, Ast_helper.Exp.ident ~loc:field.loc  field)
+      Some (field, removeModuleNameFromPunnedFieldValue (Ast_helper.Exp.ident ~loc:field.loc field))
```
IwanKaramazow, line 2954: I'm wondering if it would make sense to only run the `removeModuleNameFromPunnedFieldValue` name if we started from the `Uident` branch. What do you think?

src/res_core.ml
cknitt: Done, thanks for the feedback!

src/res_core.ml
cknitt: Done

cknitt: Oops, I see there is still something wrong here.

cknitt: Now it looks good.

IwanKaramazow (approved):
Great work @cknitt

PR #483, opened 2022-04-18
Try to switch to ocaml/setup-ocaml@v2

IwanKaramazow:
There's an error on GH actions:
```
Refusing write access to /Users/runner/.opam, which is more recent than this version of opam (2.1 > 2.0), aborting.
```
Switching to `ocaml/setup-ocaml@v2` might fix it


cristianoc:
Done here:
https://github.com/rescript-lang/syntax/pull/485

PR #487, opened 2022-05-07
jsx: allow locally abstract types and type constraints on @react.component definitions

glennsl:


IwanKaramazow (approved):
👍

tests/ppx/react/expected/newtype.res.txt
```diff
@@ -0,0 +1,6 @@
+@obj external makeProps: (~a: a, ~b: 'b, ~key: string=?, unit) => {"a": a, "b": 'b} = ""
```
cristianoc, line 1: Note this won't compile: type `a`.

cristianoc:
This one won't compile:
```rescript
@react.component
let make = (type a, ~a: a, ~b, _) => <div />
```

gives


```
FAILED: src/Demo.cmj

  We've found a bug for you!
  /Users/cristianocalcagno/GitHub/rescript-compiler/foo/src/Demo.res:5:25

  3 │ 
  4 │ @react.component
  5 │ let make = (type a, ~a: a, ~b, _) => <div />
  6 │  
  7 │ 

  This type constructor, `a`, can't be found.

```

tests/ppx/react/expected/newtype.res.txt
IwanKaramazow: Yes, the transform is incorrect. Will revert.

IwanKaramazow: @glennsl reverted the PR for now, we'll need a new PR with a fix.

glennsl: Crap. Shouldn't have assumed the tests were run through the compiler. I'll fix. I suspect this will make it a much more invasive change though.

IwanKaramazow: Do we also have a concrete piece of UI code to verify the result?

glennsl: I can certainly make one. Is there anywhere I can put this, or do you just want something to test manually on the side?

IwanKaramazow: Just a gist is ok, we don't have "full stack integration tests", so manual it is.

PR #490, opened 2022-05-08
jsx: allow locally abstract types and type constraints on @react.component

glennsl:
This is a continuation of #487 that fixes the type annotations for locally abstract types. Because locally abstract types can't be defined in type annotations, they have to be replaced with plain type variables, and this has to be done recursively since they can occur anywhere in a type.

The only limitation I have intentionally imposed here is that extensions are not allowed in type annotations when locally abstract types are used, because that would require traversal of the entire AST, which is no fun.

As requested, here's a complete example:

```js
module type Show = {
  type t
  let show: t => string
}

module Int = {
  type t = int
  let show = string_of_int
}

module Thing = {
  @react.component
  let make = (type a, ~model as module(Model: Show with type t = a), ~value: a) =>
      <div> {value->Model.show->React.string} </div>
}

let make = () =>
  <Thing model=module(Int) value=42 />
```

I hope this one covers all the bases :)

### DIFF
--- a/src/reactjs_jsx_ppx_v3.ml
+++ b/src/reactjs_jsx_ppx_v3.ml
@@ -233,43 +233,15 @@ let makeExternalDecl fnName loc namedArgListWithKeyAndRef namedTypeList =
     (makePropsType ~loc namedTypeList)
   [@@raises Invalid_argument]
 
-let rec newtypeToVar newtype typ =
-  let traverse = newtypeToVar newtype in
-  { typ with ptyp_desc =
+let newtypeToVar newtype type_ =
+  let typ (mapper : Ast_mapper.mapper) typ =
     match typ.ptyp_desc with
-    | Ptyp_constr ({ txt = Lident name }, _) when name = newtype -> Ptyp_var newtype
-    | Ptyp_constr (ident, args) -> Ptyp_constr (ident, List.map traverse args)
-    | Ptyp_arrow (label, typ, rest) -> Ptyp_arrow (label, traverse typ, traverse rest)
-    | Ptyp_tuple (types) -> Ptyp_tuple (List.map traverse types)
-    | Ptyp_class (ident, args) -> Ptyp_class (ident, List.map traverse args)
-    | Ptyp_alias (typ, alias) -> Ptyp_alias (traverse typ, alias)
-    | Ptyp_poly (vars, rest) -> Ptyp_poly (vars, traverse rest)
-    | Ptyp_variant (fields, flag, labels) ->
-      let fields =
-        List.map (function
-          | Rtag (label, attrs, flag, args) -> Rtag (label, attrs, flag,  List.map traverse args)
-          | Rinherit typ -> Rinherit (traverse typ))
-          fields
-      in
-      Ptyp_variant (fields, flag, labels)
-    | Ptyp_object (fields, flag) ->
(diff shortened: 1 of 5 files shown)


src/reactjs_jsx_ppx_v3.ml
```diff
@@ -232,6 +233,44 @@ let makeExternalDecl fnName loc namedArgListWithKeyAndRef namedTypeList =
     (makePropsType ~loc namedTypeList)
   [@@raises Invalid_argument]
 
+let rec newtypeToVar newtype typ =
+  let traverse = newtypeToVar newtype in
+  { typ with ptyp_desc =
+    match typ.ptyp_desc with
```
cristianoc, line 239: How about using `Ast_mapper`. That should take care of most boilerplate.

glennsl: I'm not really sure how to run a mapper on a single type expression, or even a general expression. Looks to me like it only does source to source transformation. But if possible, that would be ideal I think.

cristianoc: Looks like this:

```ml
let newtypeToVar newtype type_ =
  let typ (mapper : Ast_mapper.mapper) typ =
    match typ.ptyp_desc with
    | Ptyp_constr ({txt = Lident name}, _) when name = newtype ->
      {typ with ptyp_desc = Ptyp_var newtype}
    | Ptyp_extension _ ->
      Location.raise_errorf
        "extensions are not allowed in @react.component definitions"
    | _ -> Ast_mapper.default_mapper.typ mapper typ
  in
  let mapper = {Ast_mapper.default_mapper with typ} in
  mapper.typ mapper type_
```

cristianoc: Another questions is what happens if the original type contains both `a` and `'a`, worth trying to construct such an example to see what happens, that there is no accidental capture.

glennsl: Ah, thanks for the example. Makes much more sense now! Is it still necessary to raise an error with exceptions? I would think the mapper would handle that according to expectations now.

cristianoc: Yes it would. Just wanted to keep the same semantics in the example.

cristianoc: To illustrate the ast mapper.

glennsl: > Another questions is what happens if the original type contains both a and 'a, worth trying to construct such an example to see what happens, that there is no accidental capture.

Yep, that does happen. Inferred unification variables won't collide, if there's a newtype `a` it will avoid using `'a` for unification variables, but if there's an explicit constraint on `'a` it will collide.

I can think of two ways to address that:
1. Mangle the variable in some way, which makes conflicts much less likely, but not impossible and I worry that it might make errors confusing.
2. Do another pass first to find all type variables in order to pick one that isn't already used.

glennsl: Seems rescript will accept anything these days with `\""`, but definitely makes it more unlikely that there's a conflict at least! And I think using a name like `type-a` will make it pretty obvious what it refers to.

cristianoc (approved):

PR #492, opened 2022-05-13
WIP: Run CI build for macOS, Linux and Windows

cknitt:


IwanKaramazow:
@cknitt thanks for the initiative, but running the tests on an additional platform(s) won’t test anything meaningful. The compiler’s integration tests should cover it.

cknitt:
Ok, no problem.

This means that if/when we merge the syntax repo into the rescript-compiler repo, the parser tests will only be run for macOS, too, whereas we aim to run the other tests in the rescript-compiler repo for all three platforms?

cknitt:
@IwanKaramazow One benefit of building for multiple platforms would also be that we could automatically provide the built `rescript.exe` as an artifact for each platform on each commit.

cknitt:
Done with #505.

PR #494, opened 2022-05-18
Fix missing js attribute for string in braces

### LINKED ISSUE #466: Possibly forgotten change of string quotation delimiter
There is an oddity in parsing strings after merging the pull #433.

`"string"` is parsed as `{js|string|js}`
while
`{"string"}` is parsed as `"string"` (with `ns.braces` attribute)

https://github.com/rescript-lang/syntax/blob/d9c44bb22892dc4e77334f064d8bccd119fecde9/src/res_core.ml#L2749

cknitt:
Fix #466


cknitt:
@IwanKaramazow rebased after removing refmt, builds now.

IwanKaramazow (approved):
Great 👍

PR #498, opened 2022-05-26
Remove refmt

cknitt:


cristianoc:
Thanks for the cleanup!
@IwanKaramazow with all those idempotency tests gone, is something lost (except for re support)?

IwanKaramazow:
To be safe we should reformat all Reason files to ReScript and keep them in the tests

cknitt:
👍 Will do.

cristianoc:
```
  Warning Dead Value
  File "./src/res_driver_reason_binary.mli", line 4, characters 0-93
  +extractConcreteSyntax is never used
  <-- line 4
    string -> Res_token.Comment.t list * (string * Location.t) list [@@dead "+extractConcreteSyntax"] 

  Exception Analysis
  File "./src/res_multi_printer.ml", line 53, characters 4-9
  print might raise exit (res_multi_printer.ml:59:12) and is annotated with redundant @raises Sys_error
```

So it looks like `extractConcreteSyntax` has become unused now.
To check: is it used by the compiler repo?

cristianoc:
Probably a good idea to build the compiler starting from this commit and check that all builds.
If all is good, then `extractConcreteSyntax` can be removed.

cknitt:
@cristianoc It seems `extractConcreteSyntax` is not used in the compiler.

The compiler builds fine with this version of the syntax repo - if I comment out https://github.com/rescript-lang/rescript-compiler/blob/57a169198bd824dd7c971bc09e2867226fb14937/jscomp/main/rescript_compiler_main.ml#L186.

PR #501, opened 2022-05-26
Remove unused src/406

cknitt:


cristianoc:
As an aside, those stdlib files are not exactly those that will need to be vendored later.
Those to be vendored should reflect the simplification in the current compiler w.r.t. stock OCaml (no classes), and be compilable with a 4.08+ compiler. As other repos do.
So these can go.

Issue #499, opened 2022-05-26
Parsing and printing pipe expressions with attributes


cristianoc:
See https://github.com/rescript-association/reanalyze/issues/164

1 `(@doesNotRaise [])->Belt.Array.getExn(0)` formats to `@doesNotRaise []->Belt.Array.getExn(0)` so this is annotating the argument not the result
2 `@doesNotRaise ([]->Belt.Array.getExn(0))` formats to `[]->Belt.Array.getExn(0)` so it completely destroys the annotation.

cristianoc:
This should be a combination of the syntax and ppx for pipe (which is in the compiler).

PR #503, opened 2022-05-27
Fix printing of parens in binary expressions with attributes

IwanKaramazow:
Parens should be printed:
```rescript
(@doesNotRaise [])->Belt.Array.getExn(0)
```

Fixes https://github.com/rescript-lang/syntax/issues/499


cristianoc (approved):
Sweet!

PR #504, opened 2022-05-27
Printable attribute improvements

cknitt:
* Spelling fix: printeable -> printable
* Avoid unnecessary List construction
* Make sure filterPrintableAttributes and partitionPrintableAttributes use the same definition of "printable"


IwanKaramazow (approved):
Thank you so much for this. This is great.

PR #510, opened 2022-05-29
441: fix missing comments in switch

cknitt:


IwanKaramazow (approved):
Thanks, this is good for now.

When we'll support comments around/in braced expressions, we'll probably have to tweak this again to take the location of the braces in mind.

PR #511, opened 2022-05-29
fix outcome printer bugs

glennsl:
I took the existing expected test result, pasted it into the rescript playground, fixed all the errors there, then fixed the code to actually produce that result. A pretty rough process, but the fixes were all pretty simple so I hope this is accurate enough.

### DIFF
--- a/src/res_outcome_printer.ml
+++ b/src/res_outcome_printer.ml
@@ -350,7 +350,7 @@ let printPolyVarIdent txt =
         let i = ref 0 in
         let package = Doc.join ~sep:Doc.line ((List.map2 [@doesNotRaise]) (fun lbl typ ->
           Doc.concat [
-            Doc.text (if i.contents > 0 then "and " else "with ");
+            Doc.text (if i.contents > 0 then "and type " else "with type ");
             Doc.text lbl;
             Doc.text " = ";
             printOutTypeDoc typ;
--- a/src/res_outcome_printer.ml
+++ b/src/res_outcome_printer.ml
@@ -233,9 +233,11 @@ let printPolyVarIdent txt =
        )
      | Otyp_alias (typ, aliasTxt) ->
        Doc.concat [
+         Doc.lparen;
          printOutTypeDoc typ;
          Doc.text " as '";
-         Doc.text aliasTxt
+         Doc.text aliasTxt;
+         Doc.rparen
        ]
      | Otyp_constr (
         Oide_dot (Oide_dot (Oide_ident "Js", "Fn") , "arity0"), (* Js.Fn.arity0 *)
--- a/src/res_outcome_printer.ml
+++ b/src/res_outcome_printer.ml
@@ -243,7 +243,7 @@ let printPolyVarIdent txt =
        ) ->
         (* Js.Fn.arity0<t> -> (.) => t *)
         Doc.concat [
-          Doc.text "(.) => ";
+          Doc.text "(. ()) => ";
           printOutTypeDoc typ;
         ]
      | Otyp_constr (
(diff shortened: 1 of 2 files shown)


cristianoc:
This is great, we can use this immediately in the editor extension (which vendors the outcome printer).
In particular, the arity 0 bug is very visible there on hover.

For the alias type, would you add this example:
```rescript
let alias1: [> rgb] as 'a
let alias2: ([> rgb] as 'a)
```
For the deciding when to add the parens, there should be same analogous logic in the normal pretty printer for signatures.

src/res_outcome_printer.ml
```diff
@@ -233,17 +233,19 @@ let printPolyVarIdent txt =
...
+         Doc.rparen
        ]
      | Otyp_constr (
         Oide_dot (Oide_dot (Oide_ident "Js", "Fn") , "arity0"), (* Js.Fn.arity0 *)
         [Otyp_constr (Oide_ident ident, [])] (* int or unit or string *)
        ) ->
         (* Js.Fn.arity0<int> -> (.) => int*)
         Doc.concat [
-          Doc.text "(.) => ";
+          Doc.text "(. ()) => ";
```
IwanKaramazow, line 248: I wonder if we should also support `(.) => int` when parsing type expressions. We do support it in non-type expressions.

cristianoc: I did not know it was not supported. Sounds like it should be, to avoid confusion.

glennsl: > For the alias type, would you add this example:
> 
> ```rescript
> let alias1: [> rgb] as 'a
> let alias2: ([> rgb] as 'a)
> ```

Not entirely sure what you mean here. I can add something like this to `test/oprint/oprint.res`:

```rescript
let alias1: [> rgb] as 'a = #Red
let alias2: ([> rgb] as 'a) = #Green
```

but that wouldn't cause the outcome printer to produce an alias, just `[> rgb]`.

glennsl:
> For the alias type, would you add this example:
> 
> ```rescript
> let alias1: [> rgb] as 'a
> let alias2: ([> rgb] as 'a)
> ```

Not entirely sure what you mean here. I can add something like this to `test/oprint/oprint.res`:

```rescript
let alias1: [> rgb] as 'a = #Red
let alias2: ([> rgb] as 'a) = #Green
```

but that wouldn't cause the outcome printer to produce an alias, just `[> rgb]`.

src/res_outcome_printer.ml
glennsl: This surprised me too, and I think it should.

But also, for the record, `refmt` will convert `(.)` into `(. ())` for value expressions and `(. ())` into `(. unit)` for type expressions. Perhaps we should run this through `refmt` to ensure it produces idiomatic code too?

cristianoc:
@glennsl thoughts about adding parsing for `(.) => int` ?

cristianoc:
Also, would you rebase on master.

glennsl:
> @glennsl thoughts about adding parsing for e.g. (.) => int ?

You mean beyond agreeing that it's a good idea?

> Also, would you rebase on master.

Done, but seems I can no longer run tests on Linux?

cristianoc:
> > @glennsl thoughts about adding parsing for e.g. (.) => int ?
> 
> You mean beyond agreeing that it's a good idea?
> 
> > Also, would you rebase on master.
> 
> Done, but seems I can no longer run tests on Linux? (Edit: I can run the commands manually though, just skipping reanalyze)

opam install reanalyze, does it work for you?

cristianoc:
Added an issue to track parsing `(.) => ...` in type expressions add https://github.com/rescript-lang/syntax/issues/549

I guess this can be merged.

glennsl:
> opam install reanalyze, does it work for you?

Installing it works, but running it does not. Some kind of CMI parse error being raised I think. The README says "tests will currently run on macOS only", so I figured this was known.

cknitt:
The README is outdated, I will update it. Current status is that all tests are working on all platforms, except for roundtrip tests on Windows.

PR #514, opened 2022-06-01
Fix example

cknitt:
Fix example in README so that it actually compiles. Well, except for the `FS.readFile` which I guess is pseudocode?


IwanKaramazow:
> Well, except for the FS.readFile which I guess is pseudocode?

Yes, reading a file with Ocaml makes you reach StackOverflow

PR #524, opened 2022-06-06
Vendor 4.06 compiler libs and support host compiler from 4.08 onwards.

cknitt:
Current status:

Copied the compiler libs from https://github.com/rescript-lang/rescript-vscode/tree/master/analysis/vendor/compiler-libs-406.

Everything is compiling fine - except for the tests. 😞

When I try to add the missing dependencies for `res_test.ml` (in particular, `Typemod` and its dependencies) by copying them from the compiler sources, I am not really getting anywhere. It feels like

> "You wanted a banana but what you got was a gorilla holding the banana and the entire jungle."

I am afraid I will need some more guidance here, @cristianoc.


cristianoc:
Looking at that. Indeed there are tricky bits.

cristianoc:
@cknitt OK now it compiles, and crashes when testing.
The crash is because it can't open module `Pervasives`:
```
    open_implicit_module "Pervasives" initial
```
I guess this is normally done by packaging the compiler, and loading it at startup. So not sure how it would work in this setting.

cristianoc:
Basically, the tests reach further into the compiler than the other tools. So this is a bit of an unexplored part.

cristianoc:
++ @IwanKaramazow any ideas?

cristianoc:
Just realised all this just for a single file: `tests/oprint/oprint.res`.
Definitely the test setup should be simplified.

cristianoc:
Changed the outcome printer test to not use anything from Pervasives.
Also restored reanalyze.
So now tests run as before.

cristianoc:
Just a reminder of an offline discussion. Tests on linux CI fail because `zsh` is not present. They should succeed once `bash` is used for `test.sh` instead.

cristianoc:
Hopefully Windows will find the new published reanalyze soon.

README.md
```diff
@@ -23,14 +23,12 @@ J. Blow.
 
 Required:
 
-- OCaml 4.06.1
+- OCaml 4.08 or later
```
glennsl, line 26: FYI: I could not get master to build now with 4.08.1. 4.13.1 worked though.

cristianoc: For me 4.08 does not show anymore in the available switches, and 4.10 works.
Probably some feature introduced later is used, which is OK.
No plan to support specific old versions. Will just update the README.

PR #534, opened 2022-06-10
Printer: auto-pun record fields

cknitt:
In #480, I already prevented the printer from expanding `{a, b} to {a: a, b: b}`.

At the time, I refrained from making it collapse `{a: a, b: b}` to `{a, b}` like it used to be the case in Reason.
But I think it would actually be the better behavior to always print record fields as punned when possible.

Now that the next release is going to be a major version (10.0), it would IMHO be fine to make that change.

### DIFF
--- a/src/res_printer.ml
+++ b/src/res_printer.ml
@@ -4826,8 +4826,7 @@ and printRecordRow (lbl, expr) cmtTbl punningAllowed =
     match expr.pexp_desc with
     | Pexp_ident({txt = Lident key; loc = keyLoc}) when (
       punningAllowed &&
-      Longident.last lbl.txt = key &&
-      lbl.loc.loc_start.pos_cnum == keyLoc.loc_start.pos_cnum
+      Longident.last lbl.txt = key
     ) ->
       (* print punned field *)
       printLidentPath lbl cmtTbl;
--- a/tests/conversion/reason/expected/bracedJsx.res.txt
+++ b/tests/conversion/reason/expected/bracedJsx.res.txt
@@ -91,7 +91,7 @@ let make = () => {
           ],
         ),
       }
-    | SetValue(input) => {...state, input: input}
+    | SetValue(input) => {...state, input}
     }
   , {history: [], input: ""})
 
--- a/tests/conversion/reason/expected/braces.res.txt
+++ b/tests/conversion/reason/expected/braces.res.txt
@@ -20,5 +20,5 @@ let getDailyNewCases = x =>
   | Pair({prevRecord, record}) =>
     let confirmed = record.confirmed - prevRecord.confirmed
     let deaths = record.deaths - prevRecord.deaths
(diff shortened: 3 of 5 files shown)


cristianoc:
What's the corresponding behaviour for types? Being consistent there could be positive, all else equal.

cknitt:
@cristianoc @IwanKaramazow  Some history: Reason used to do this for both record values and types, i.e.

```reason
type a = int;
type b = string;

type c = { a: a, b: b };

let a = 1;
let b = "";

let c = { a: a, b: b };
```

would reformat to

```js
type a = int;
type b = string;

type c = {
  a,
  b,
};

let a = 1;
let b = "";

let c = {a, b};
```

ReScript (up to the last release 9.1.4) used to do the inverse - change `{a, b}` to `{a: a, b: b}` for both values and types. I always felt this was a regression, as the shorter form seems much more readable to me. Much less "visual noise" IMHO, especially for records with many fields and/or long field names.

Current state in ReScript after #480 is that record values are printed exactly the way the user wrote them (whether punned or not), but record types are still expanded. The latter is because Cheng and Bob stated in https://github.com/rescript-lang/syntax/pull/64#issuecomment-663914884 and https://github.com/rescript-lang/syntax/pull/64#issuecomment-666879913 that they didn't want punning for record type declarations.

I do not care that much about the type declarations, but I strongly favor the punned version for record values. It is nice that, since #480, the printer is leaving the punned version untouched, but I would prefer if the punned version was actually enforced by the printer to ensure a consistent and more readable style in ReScript codebases.

cristianoc:
It seems for each combination one could come up with some argument. Can't see compelling reasons either way.

cristianoc:
Wondering how relevant the question for types is. Most often the names would be different.

cristianoc:
@cknitt any thoughts on how to progress this? Perhaps more feedback from the forum?

cknitt:
Yes, will ask on the forum later today. 👍

IwanKaramazow:
One thought is that for the type counterparts, punning might be pretty confusing:
```rescript
type coordinate = {x, y: int}
```
One could argue that some people will interpret `x` as having type `int` and not `x`.

cristianoc:
> One thought is that for the type counterparts, punning might be pretty confusing:
> 
> ```rescript
> type coordinate = {x, y: int}
> ```
> 
> One could argue that some people will interpret `x` as having type `int` and not `x`.

Remember that discussion. Yes symmetry between types and values does not seem like a good idea.

cristianoc:
Looks like this can be merged, based on discussions in the forum.

cristianoc (approved):

cristianoc:
Leaving a link to the discussion: https://forum.rescript-lang.org/t/printing-punned-record-fields/3447

cknitt:
I have added test cases for single-element records (which cannot be punned).

I will now close the discussion and merge the PR.

PR #542, opened 2022-06-14
Do not parse past Eof

cristianoc:
Parsing past Eof can lead to infinite loops, as it violates the assumptions of the termination checker.
See: https://github.com/rescript-lang/syntax/pull/540

This PR makes `Parser.next` assert false when called on Eof. This should not happen as one should check the token before calling `Parser.next`.

The one exception is during lookahead, for which we provide a `nextUnsafe` function which does not make progress.

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -259,7 +259,7 @@ let isEs6ArrowExpression ~inTernary p =
            * *)
           false
         | _ ->
-          Parser.next state;
+          Parser.nextUnsafe state;
           (* error recovery, peek at the next token,
            * (elements, providerId] => {
            *  in the example above, we have an unbalanced ] here
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -623,7 +623,7 @@ let rec parseLident p =
       None
     ) else (
       let rec loop p =
-        if not (Recover.shouldAbortListParse p)
+        if not (Recover.shouldAbortListParse p) && p.token <> Eof
         then begin
           Parser.next p;
           loop p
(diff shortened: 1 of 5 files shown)


cristianoc:
This turns potential infinite loops into runtime exceptions when calling `Parser.next`.
I wad editing the code for exceptions, and realised there are a few parts where there is very nearly an infinite loop (it's an innocent-looking code change away). Which after this change manifests as a runtime exception.
In here https://github.com/rescript-lang/syntax/pull/543 the exception analysis is used to make sure this does not happen.

cristianoc:
Summarising: preventing infinite loops arising from cases where one keeps on reading `Eof` over and over again statically is a bit tricky right now.
But, it's fairly easy to detect when this happens. The current PR does just that: detection. This will only fire when there's another mistake in the implementation of the parser.
Currently, detection gives an assert false, but perhaps it makes sense to give dedicated error instead. Notice that recovery is not an option in this case (unless one were to use effect handlers in the implementation).

cristianoc:
At the same time the result of https://github.com/rescript-lang/syntax/pull/543 can be used to see where the logic had to be modified, and try to use that to construct infinite loop repros.

Here is one:

```rescript
let foo = x =>
  switch x {
  | `${
```

cristianoc:
Another case to investigate: https://github.com/rescript-lang/syntax/pull/543/files#r897400872

cristianoc:
Another one: https://github.com/rescript-lang/syntax/pull/543/files#r897402749

cristianoc:
The last unclear case found during audit does not seem a real infinite loop.
So this should cover all cases considered.

cristianoc:
Added an issue to track better error messages: https://github.com/rescript-lang/syntax/issues/550

PR #543, opened 2022-06-14
Eof audit (DO NOT MERGE)

cristianoc:
There are 2 functions that the parser can call:
- next can raise Eof
- nextUnsafe does not raise

The former is counted in the progress measure. It cannot be part of infinite loops, as eventually Eof would be reached and the exception thrown.

This means that every use of Parser.next needs to be audited for exceptions, since infinite loops have been turned into runtime exceptions.
Sometimes it's enough to use Parser.expect if we know exactly what is the current token.
Sometimes the current token has just been checked, so a `@doesNotRaise` annotation is added.
Sometimes there are several call-sites to the current function that have different invariants, and this is where some changes were required.

Also audited `expect`. There are 2 variants:
- `expect`: use when it's clear the token is not Eof
- `expectUnsafe`: when it's unclear whether the token is Eof


src/res_core.ml
```diff
@@ -640,23 +642,24 @@ let rec parseLident p =
...
   | Lident ident ->
-    Parser.next p;
+    Parser.next p [@doesNotRaise];
     let loc = mkLoc startPos p.prevEndPos in
     (ident, loc)
+  | Eof ->
+    ("_", mkLoc startPos p.prevEndPos)
   | _ ->
-    begin match recoverLident p with
+    begin match recoverLidentNoEof p with
```
cristianoc, line 651: OK to call: we just checked for Eof in the previous case of the match.

src/res_core.ml
```diff
@@ -640,23 +642,24 @@ let rec parseLident p =
...
     | Some () ->
       parseLident p
     | None ->
       ("_", mkLoc startPos p.prevEndPos)
     end
 
-let parseIdent ~msg ~startPos p =
+let parseIdentNoNext ~msg ~startPos p =
```
cristianoc, line 658: We do `Parser.next` after calling this function.
This is because the caller has more context on what the token can be.

src/res_core.ml
```diff
@@ -665,39 +668,42 @@ let parseIdent ~msg ~startPos p =
...
-    Parser.next p;
     (tokenTxt, mkLoc startPos p.prevEndPos)
   | _token ->
     Parser.err ~startPos p (Diagnostics.message msg);
-    Parser.next p;
     ("", mkLoc startPos p.prevEndPos)
 
+let parseIdent ~msg ~startPos p =
+  let res = parseIdentNoNext ~msg ~startPos p in
+  Parser.nextUnsafe p;
```
cristianoc, line 678: When we don't know, just do `nextUnsafe`. That's OK as long as termination analysis is happy.

src/res_core.ml
```diff
@@ -710,21 +716,33 @@ let parseValuePath p =
...
+        Longident.Lident ident
+      )
+    | token ->
+      Parser.err p (Diagnostics.unexpected token p.breadcrumbs);
+      Longident.Lident "_"
   in
-  Parser.next p;
+  ident
+
+let parseValuePath p =
```
cristianoc, line 736: This is what we call when we don't know what the token is.

src/res_core.ml
```diff
@@ -710,21 +716,33 @@ let parseValuePath p =
...
+  let startPos = p.Parser.startPos in
+  let ident = parseValuePathNoNext p in
+  Parser.nextUnsafe p;
+  Location.mkloc ident (mkLoc startPos p.prevEndPos)
+
+let parseValuePathNotEof p = (* Only call when current token is not Eof *)
```
cristianoc, line 742: This is what we call when we know the token is not Eof.

src/res_core.ml
```diff
@@ -1374,11 +1396,10 @@ and parseConstrainedPatternRegion p =
  *	 | field , _
  *	 | field , _,
  *)
-and parseRecordPatternField p =
-  let label = parseValuePath p in
+and parseRecordPatternField ~(label: Longident.t Location.loc) p =
```
cristianoc, line 1399: We do the label thing in the caller, which has more context on what the token is.

src/res_core.ml
```diff
@@ -1391,12 +1412,12 @@ and parseRecordPatternField p =
 and parseRecordPatternItem p =
   match p.Parser.token with
   | DotDotDot ->
-    Parser.next p;
-    Some (true, PatField (parseRecordPatternField p))
+    Parser.expect DotDotDot p;
+    Some (true, PatField (parseRecordPatternField ~label:(parseValuePath p) p))
```
cristianoc, line 1416: Here we *don't* know what the current token is.
But, termination analysis is happy so we're good.

src/res_core.ml
```diff
@@ -1391,12 +1412,12 @@ and parseRecordPatternField p =
...
   | Uident _ | Lident _ ->
-    Some (false, PatField (parseRecordPatternField p))
+    Some (false, PatField (parseRecordPatternField ~label:(parseValuePathNotEof p) p))
```
cristianoc, line 1418: Here we do know the token is not `Eof`. If we did not propagate this info, which lets us use `next` instead of `nextUnsafe` termination analysis would fire.

src/res_core.ml
```diff
@@ -1978,15 +1999,12 @@ and parseFirstClassModuleExpr ~startPos p =
     Ast_helper.Exp.pack ~loc modExpr
   end
 
-and parseBracketAccess p expr startPos =
-  Parser.leaveBreadcrumb p Grammar.ExprArrayAccess;
-  let lbracket = p.startPos in
-  Parser.next p;
-  let stringStart = p.startPos in
+and parseBracketAccess p ~lbracket expr startPos =
```
cristianoc, line 2002: We do the `next` in the caller, which has more context on what the token is.

src/res_core.ml
```diff
@@ -2077,7 +2095,10 @@ and parsePrimaryExpr ~operand ?(noCall=false) p =
         loop p (Ast_helper.Exp.field ~loc expr lident)
       end
     | Lbracket when noCall = false && p.prevEndPos.pos_lnum == p.startPos.pos_lnum ->
-      parseBracketAccess p expr startPos
+      Parser.leaveBreadcrumb p Grammar.ExprArrayAccess;
+      let lbracket = p.startPos in
+      Parser.expect Lbracket p;
```
cristianoc, line 2100: What was a `next` inside the callee, is now an `expect` in the caller.

src/res_core.ml
```diff
@@ -2964,16 +2985,16 @@ and parseRecordRow p =
...
     Parser.err p (Diagnostics.message ErrorMessages.recordExprSpread);
-    Parser.next p;
+    Parser.expect DotDotDot p;
   | _ -> ()
   in
   match p.Parser.token with
   | Lident _ | Uident _ ->
     let startToken = p.token in
-    let field = parseValuePath p in
+    let field = parseValuePathNotEof p in
```
cristianoc, line 2994: We know this is not `Eof`.

cristianoc:
This is way too noisy to even think about merging something like this.
But it serves as an indication of the gap between what's out there and what one would need in order to get stronger static guarantees.

src/res_core.ml
```diff
@@ -963,15 +984,17 @@ let parseTemplateConstant ~prefix (p : Parser.t) =
...
+    Parser.next p [@doesNotRaise];
     let txt = if p.mode = ParseForTypeChecker then parseTemplateStringLiteral txt else txt in
     Parsetree.Pconst_string (txt, prefix)
   | _ ->
     let rec skipTokens () =
-      Parser.next p;
-      match p.token with
-      | Backtick -> Parser.next p; ()
-      | _ -> skipTokens ()
+      if p.token <> Eof then (
```
cristianoc, line 992: And this is an infinite loop:
```res
let foo = x =>
  switch x {
  | `${
```

See https://github.com/rescript-lang/syntax/pull/542

src/res_core.ml
```diff
@@ -1268,26 +1291,28 @@ and skipTokensAndMaybeRetry p ~isStartOfGrammar =
...
       end
+    else if p.token = Eof then
+      None
     else
       begin
-        Parser.next p;
+        Parser.next p [@doesNotRaise];
         let rec loop p =
-          if not (Recover.shouldAbortListParse p)
+          if not (Recover.shouldAbortListParse p) && p.token <> Eof
```
cristianoc, line 1313: This looks like could be another infinite loop

src/res_core.ml
```diff
@@ -4049,46 +4075,51 @@ and parseTypeParameters p =
...
-    let endPos = p.prevEndPos in
-    let typ = List.fold_right (fun (uncurried, attrs, argLbl, typ, startPos) t ->
-      let attrs = if uncurried then uncurryAttr::attrs else attrs in
-      Ast_helper.Typ.arrow ~loc:(mkLoc startPos endPos) ~attrs argLbl typ t
-    ) parameters returnType
-    in
-    {typ with
-      ptyp_attributes = List.concat [typ.ptyp_attributes; attrs];
-      ptyp_loc = mkLoc startPos p.prevEndPos}
+    if p.token <> Eof then (
```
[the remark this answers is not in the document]
cristianoc: Does not look like it, it would need to go into `parseTypExpr`/`parseEs6ArrowType` loop, but that only works if there are fresh `=>` or `~` to look at.

PR #545, opened 2022-06-14
put parseTagName at toplevel

cristianoc:
Continuation of https://github.com/rescript-lang/syntax/pull/540


src/res_grammar.ml
```diff
@@ -361,6 +364,7 @@ let isListTerminator grammar token =
   | PackageConstraint, token when token <> And -> true
   | ConstructorDeclaration, token when token <> Bar -> true
   | AttributePayload, Rparen -> true
+  | TagNames, Rbracket -> true
```
cristianoc, line 367: This actually needs to be specified in order for `parseRegion` to work.

src/res_grammar.ml
```diff
@@ -336,6 +338,7 @@ let isListElement grammar token =
   | JsxAttribute -> isJsxAttributeStart token
   | JsFfiImport -> isJsFfiImportStart token
   | AttributePayload -> token = Lparen
+  | TagNames -> token = Hash
```
cristianoc, line 341: This is not currently used, as when returning `None` the current token won't be `Hash`, but it's good to have it here for later in case the parse tag function changes and does some more refined error recovery after `#`.

Issue #439, opened 2021-06-16
Docstring (/** */) support


ryyppy:
(Note: This is a tracking issue for our docstring `/** */` support listed in our roadmap)

Goal is to get rid of `@ocaml.doc` by introducing a dedicated `/** */` syntax.

PR #546, opened 2022-06-15
Explore doc comments.

cristianoc:
crf https://github.com/rescript-lang/syntax/issues/439


cristianoc:
This. was a bit more complex than expected -- need to revisit later.

Issue #521, opened 2022-06-04
Plan for JSX V4.


cristianoc:
With the introduction of more general structural typing https://forum.rescript-lang.org/t/ann-more-strict-checks-in-missed-labels/2117 most of the JSX ppx could be removed. In particular, one could just use functions from the language directly.

A plan would require
- [x] Describe precisely how it works. (See https://github.com/rescript-lang/syntax/pull/517/files#diff-c97c31727a92fde7cad7a258699c4d6bf67cba2654d93992c758bf94df5f8411).
- [x] Evaluate whether it is true that one can just "use the language" to define functions for components. Find out if there are issues and what these are.
- [x] Figure out a solution to the expressivity problem of `@obj` records: https://github.com/rescript-lang/syntax/pull/517#issuecomment-1151701199
- [ ] Check that hover and jump to location in the editor work.
- [ ] Check that type error messages report the desired location.
- [ ] Explore JSX preserve mode (i.e. output jsx syntax in the generated code). See https://github.com/rescript-lang/syntax/issues/539.
- [ ] Explore React's own jsx transform (https://forum.rescript-lang.org/t/jsx-v4-next-rescript-react-0-10-x/3431/10)
- [ ] Plan a migration path. And what about dependencies.
- [ ] Engage with the community for feedback. This would be a major visible change. (See [forum post](https://forum.rescript-lang.org/t/jsx-v4-next-rescript-react-0-10-x/3431))

cristianoc:
++ @rickyvetter

ryyppy:
**Things to consider (that may or may not be solved by simplifying the ppx):**
- Make sure it's easy to comply to the fast-refresh naming rules (capitalized functions, no leading `$$`) -> also see [this forum post](https://forum.rescript-lang.org/t/rescript-nextjs-template-2021-update/829/5?u=ryyppy)

(will probably edit this comment later to add more items)

PR #547, opened 2022-06-15
JSX V4

mattdamon108:
This PR rewrites the JSX PPX which is aligned to https://github.com/rescript-lang/syntax/issues/521.
Here's the [Spec](https://github.com/rescript-lang/syntax/blob/0bbb5f8e30f220c5d6fe7b836d663a7272cbedff/cli/JSXV4.md) at the time of writing this. But please check if the spec has changed as part of this PR.

A further plans

* [ ] `forwardRef` interface
* [ ] Engage the feedback from the forum
* [ ] Polish the loc properly
* [ ] Test (e.g. newtype, type constraint cases)
* [ ] Explore the surface for the new JSX transform with`react/jsx-runtime`
    * [x] Check the jsx binding in `rescript-react`. PR https://github.com/rescript-lang/rescript-react/pull/1
    * [ ] Add `jsx.Fragment` binding into the above jsx binding
    * [x] Align to the core team regarding the possibility of support for jsx and createElement both or not.
* [x] Expose `ref` only for `React.forwardRef`
* [x] `RescriptReactErrorBoundary` impl and intf are mismatching in current version of rescript-react [impl](https://github.com/rescript-lang/rescript-react/blob/16dcbd8d079c7c20f3bd48fd677dfe7d70d0d020/src/RescriptReactErrorBoundary.res) [intf](https://github.com/rescript-lang/rescript-react/blob/16dcbd8d079c7c20f3bd48fd677dfe7d70d0d020/src/RescriptReactErrorBoundary.resi)


tests/ppx/react/expected/commentAtTop.res.txt
```diff
@@ -1,10 +1,14 @@
-@obj external makeProps: (~msg: 'msg, ~key: string=?, unit) => {"msg": 'msg} = "" // test React JSX file
+type props<'msg> = {@optional key: string, @optional ref: ReactDOM.Ref.currentDomRef, msg: 'msg} // test React JSX file
```
cristianoc, line 1: I guess it's known and in the plans, but `ref` should not appear on normal examples.
Only on those using `forwardRef`.

mattdamon108: Good catch, thanks. I think it is not allowed in v3 either, even though passing the ref is working for react class component. I think it is reliable not to pass the ref unless it is for sure that it react class component. https://reactjs.org/docs/refs-and-the-dom.html#refs-and-function-components

cristianoc (commented):
A small comment on style: this looks a little strange:

```ml
let makePropsTypeParamsSig namedTypeList =
  namedTypeList
  |> List.filter_map (fun (isOptional, label, _, interiorType) ->
         if label = "key" || label = "ref" then None
         else if isOptional then Some (extractOptionalCoreType interiorType)
         else Some interiorType)
```

as it removes an `@optional ` from the type. But that annotation was added by the ppx itself.

It might indicate that the logic for doing this should probably be further back in the caller, not in this function. So that the `@optional` attribute is never added in the first place. Otherwise, it could be easy to introduce bugs via refactoring without realizing it.

cristianoc (commented):
Also this is pre-existing, but now that the code is formatted it really jumps to the eye.

Wrapping things in `let jsxMapper () =` indents everything, and groups a bunch of functions that don't really need to be grouped. How about just remove it and bring everything to the top level.

cristianoc:
> Also this is pre-existing, but now that the code is formatted it really jumps to the eye.
> 
> Wrapping things in `let jsxMapper () =` indents everything, and groups a bunch of functions that don't really need to be grouped. How about just remove it and bring everything to the top level.

At that point, this global state:
```
let nestedModules = ref []
```
should be created in `rewrite_implementation` and `rewrite_signature` and passed around to the functions that need it.

mattdamon108:
> A small comment on style: this looks a little strange:
> 
> ```ocaml
> let makePropsTypeParamsSig namedTypeList =
>   namedTypeList
>   |> List.filter_map (fun (isOptional, label, _, interiorType) ->
>          if label = "key" || label = "ref" then None
>          else if isOptional then Some (extractOptionalCoreType interiorType)
>          else Some interiorType)
> ```
> 
> as it removes an `@optional ` from the type. But that annotation was added by the ppx itself.
> 
> It might indicate that the logic for doing this should probably be further back in the caller, not in this function. So that the `@optional` attribute is never added in the first place. Otherwise, it could be easy to introduce bugs via refactoring without realizing it.

Absolutely, I totally forgot this part of generating sig, this strange code is written before you introduced the `@optional` syntax. I'll fix it.

mattdamon108:
> Also this is pre-existing, but now that the code is formatted it really jumps to the eye.
> 
> Wrapping things in `let jsxMapper () =` indents everything, and groups a bunch of functions that don't really need to be grouped. How about just remove it and bring everything to the top level.

cli/reactjs_jsx_ppx_v3.ml
```diff
@@ -604,12 +597,10 @@ let jsxMapper () =
     [@@raises Invalid_argument]
   in
 
-  let argToConcreteType types (name, loc, type_) =
+  let argToConcreteType types (name, _loc, type_) =
     match name with
     | name when isLabelled name -> (false, getLabel name, [], type_) :: types
-    | name when isOptional name ->
-      (true, getLabel name, [], Typ.constr ~loc {loc; txt = optionIdent} [type_])
-      :: types
```
mattdamon108: > A small comment on style: this looks a little strange:
> 
> ```ocaml
> let makePropsTypeParamsSig namedTypeList =
>   namedTypeList
>   |> List.filter_map (fun (isOptional, label, _, interiorType) ->
>          if label = "key" || label = "ref" then None
>          else if isOptional then Some (extractOptionalCoreType interiorType)
>          else Some interiorType)
> ```
> 
> as it removes an `@optional ` from the type. But that annotation was added by the ppx itself.
> 
> It might indicate that the logic for doing this should probably be further back in the caller, not in this function. So that the `@optional` attribute is never added in the first place. Otherwise, it could be easy to introduce bugs via refactoring without realizing it.

Fixed. I totally forgot this improvised logic at that time. Thanks!

cli/reactjs_jsx_ppx_v3.ml
```diff
@@ -317,953 +317,935 @@ let makePropsRecordTypeSig propsName loc namedTypeList =
...
+let module_binding nestedModules mapper module_binding =
+  let _ = nestedModules := module_binding.pmb_name.txt :: !nestedModules in
+  let mapped = default_mapper.module_binding mapper module_binding in
+  let _ = nestedModules := List.tl !nestedModules in
+  mapped
+  [@@raises Failure]
+
+(* TODO: some line number might still be wrong *)
+let jsxMapper nestedModules =
+  let jsxVersion = ref None in
```
mattdamon108, line 1234: I've remained the `jsxVersion` ref still, anyway what is it for? I've looked into the compiler code base too, but it seems not to have a specific job.

mattdamon108: Sounds good.

cristianoc:
Probably best to rebase on master (see https://github.com/rescript-lang/syntax/pull/565) sooner rather than later.
To avoid piling up merge conflicts.

mattdamon108:
What is the core team's plan for ppx v4 to support the new jsx transform? I found some feedbacks in forum that they want it, but there is a review not to support the new jsx transform. https://github.com/rescript-lang/rescript-react/pull/1#discussion_r558402028 I think it is an easy task to make the ppx v4 support the new jsx transform, but I have questions about it.
1. Does the ppx v4 need to support the new jsx transform? or support them both?
2. If the ppx v4 needs to support them both with user's preference, what is the interface to get a configuration value from the compiler?

cristianoc:
> What is the core team's plan for ppx v4 to support the new jsx transform? I found some feedbacks in forum that they want it, but there is a review not to support the new jsx transform. [rescript-lang/rescript-react#1 (comment)](https://github.com/rescript-lang/rescript-react/pull/1#discussion_r558402028) I think it is an easy task to make the ppx v4 support the new jsx transform, but I have questions about it.
> 
> 1. Does the ppx v4 need to support the new jsx transform? or support them both?
> 2. If the ppx v4 needs to support them both with user's preference, what is the interface to get a configuration value from the compiler?

One constraint is that the initial release will require no changes to rescript-react.
Does this constraint affect the ability to support the new jsx transform?

mattdamon108:
> One constraint is that the initial release will require no changes to rescript-react.
> Does this constraint affect the ability to support the new jsx transform?

The APIs for the new jsx transform are interopped already, but those are having attribute `@deprecated`. I think ppx v4 can use it without changes in rescript-react, but users will encounter countless warnings I guess.

cristianoc:
> > One constraint is that the initial release will require no changes to rescript-react.
> > Does this constraint affect the ability to support the new jsx transform?
> 
> The APIs for the new jsx transform are bound already https://github.com/rescript-lang/rescript-react/blob/ef325291733c8d42a53d87e8317371cbb6381011/src/React.res#L28, but those are having attribute `@deprecated`. I think ppx v4 can use it without changes in rescript-react, but users will encounter countless warnings I guess.

The PPX can issue code that silences deprecation warnings. Just like it can do for unused variables.
Another question is whether using the new jsx transform is a breaking change for existing projects?

mattdamon108:
> Another question is whether using the new jsx transform is a breaking change for existing projects?

Sure, the first case that comes to mind is the project which has the dependencies under React v17. But I can't claim other cases for now. Maybe we can ask it in the forum.

cristianoc:
Definitely something for the forum. Really need to know who this will break. As that would make it necessary to have it configurable.
If configurable, probably in `bsconfig.json`.

mattdamon108:
> If configurable, probably in bsconfig.json.

Yes, it is workable as one of configurations in `bsconfig.json`.
```json
"reason": {
  "react-jsx": 4,
  "react-runtime": "automatic"  // or "classic" like the Babel configuration
}
```

mattdamon108:
I'm trying to figure out how to access the bsconfig value, but I think I'm not on right track. Can you give me a clue?
AFAIK, all syntax modules are copied to `/jscomp/napkin` in compiler for building, then `napkin.cmx` is linked to `bsc.exe`. But, the parser for the `bsconfig.json` and config values are accessed in `rescript.exe` only. Is it possible for ppx module to access the config value? I guess it is only possible backward. I mean `bsc.exe` is executed with args, so that the additional cli arg is needed to be executed by `rescript.exe`.

cristianoc:
I think that's right. bsc needs to be as fast as possible so it gets all the options passed to

mattdamon108:
It makes sense that the separate binary seems an architectural choice for performance.

mattdamon108:
I figured out and tested adding a new cli arg between rescript.exe and bsc.exe successfully. Now I can get a configuration value from bsconfig.json to syntax. I have a question and your thought about user interface and API between compiler and syntax.
#### bsconfig.json
```json
"reason": {
  "react-jsx": 3 // or 4
  "react-runtime": "classic" // or "automatic"
}
```
#### cli args
```shell
-bs-jsx [3 or 4]
-bs-react-runtime [classic or automatic]
```
In this regard, If we put the ppx v3 and v4 both, then we can make users migrate their projects gracefully. What do you think?

cristianoc:
> I figured out and tested adding a new cli arg between rescript.exe and bsc.exe successfully. Now I can get a configuration value from bsconfig.json to syntax. I have a question and ask your thought about user interface and API between compiler and syntax.
> 
> #### bsconfig.json
> ```json
> "reason": {
>   "react-jsx": 3 // or 4
>   "react-runtime": "classic" // or "automatic"
> }
> ```
> 
> #### cli args
> ```shell
> -bs-jsx [3 or 4]
> -bs-react-runtime [classic or automatic]
> ```
> 
> In this regard, If we put the ppx v3 and v4 both, then we can make users migrate their projects gracefully. What do you think?

Suggestion: `"reason"` called `"react"` instead.
Then just `"jsx"` and `"runtime"`.
There will be the schema to update when this is final, but there's time for that.

This could be a good time to begin thinking about migration. Having a version flag is useful, though not enough, to support migration. The flag is what's going to be used at the *end* of a gradual migration. What's left is supporting intermediate migration steps when one only migrates e.g. one component, or one dependent project.

cristianoc:
Something to check, once the ppx version flag is in place: what happens with dependencies? I guess they will be recompiled with whatever setting is used in the current project. But one needs to verify that this is true.

cristianoc:
Btw looks like this is there already in. `rescript_compiler_main.ml`:
```ml
    "-bs-jsx", string_call (fun i -> 
        (if i <> "3" then Bsc_args.bad_arg (" Not supported jsx version : " ^  i));
        Js_config.jsx_version := 3),

```

mattdamon108:
> Btw looks like this is there already in. `rescript_compiler_main.ml`:
> 
> ```ocaml
>     "-bs-jsx", string_call (fun i -> 
>         (if i <> "3" then Bsc_args.bad_arg (" Not supported jsx version : " ^  i));
>         Js_config.jsx_version := 3),
> ```

Yes. Sorry for making a confusion. "react-jsx" and "-bs-jsx" are already there, new things are "react-runtime" and "-bs-react-runtime".

mattdamon108:
> Suggestion: "reason" called "react" instead.
> Then just "jsx" and "runtime".
> There will be the schema to update when this is final, but there's time for that.

I'm afraid that these changes in bsconfig.json will break the existing projects. It seems totally depends on the decision of the core team when introducing V4 to users. How the users migrate their projects to V4 with or without changing `bsconfig.json`. IMHO, How about leaving the existing field as they are, and just adding `runtime` only.

Or, we can add the brand new field additionally to keep the backward compatibilty.
```json
"reason": {
  "react-jsx": 3
},
"react": {
  "jsx": 4,
  "runtime": "classic"
}
```

cristianoc:
> > Suggestion: "reason" called "react" instead.
> > Then just "jsx" and "runtime".
> > There will be the schema to update when this is final, but there's time for that.
> 
> I'm afraid that these changes in bsconfig.json will break the existing projects. It seems totally depends on the decision of the core team when introducing V4 to users. How the users migrate their projects to V4 with or without changing `bsconfig.json`. IMHO, How about leaving the existing field as they are, and just adding `runtime` only.
> 
> Or, we can add the brand new field additionally to keep the backward compatibilty.
> 
> ```json
> "reason": {
>   "react-jsx": 3
> },
> "react": {
>   "jsx": 4,
>   "runtime": "classic"
> }
> ```

I would not worry about backwards compatibility as the `jsx` config is currently deprecated and not used in the compiler. As there's only 1 ppx on offer at the moment.

mattdamon108:
> Something to check, once the ppx version flag is in place: what happens with dependencies? I guess they will be recompiled with whatever setting is used in the current project. But one needs to verify that this is true.

100%, it will be recompiled. I already mentioned as one of task in this PR. `RescriptReactErrorBoundary` will be broken with V4 because its impl and intf are mismatched already. The impl doesn't use `@react.component` instead of intf.

cristianoc:
The jsx runtime could be a good opportunity to think about compiler-provided jsx types as in https://forum.rescript-lang.org/t/change-jsx-output-from-react-to-solidjs/1930/40?u=cristianoc

cristianoc:
>`RescriptReactErrorBoundary` will be broken with V4 because its impl and intf are mismatched already. The impl doesn't use `@react.component` instead of intf. I found it when I tried to build my project with the compiler v10.

Would sending a fix to rescript-react help a little bit?

mattdamon108:
> Would sending a fix to rescript-react help a little bit?

Absolutely, no problem. I'll make a PR.

mattdamon108:
> Actually maybe it's better to put the config options under `"jsx"` as it's not only `react` one wants to support. So `"jsx"` with `version` and `runtime`.

To make sure your intention clearly, is it correct? Leaving the existing "reason", and adding "jsx"?
```json
"reason": {
  "react-jsx": 3
},
"jsx": {
  "version": 4,
  "runtime": "automatic"
} 
```

cristianoc:
> > Actually maybe it's better to put the config options under `"jsx"` as it's not only `react` one wants to support. So `"jsx"` with `version` and `runtime`.
> 
> To make sure your intention clearly, is it correct? Leaving the existing "reason" and adding "jsx" or replacing it?
> 
> ```json
> "reason": {
>   "react-jsx": 3
> },
> "jsx": {
>   "version": 4,
>   "runtime": "automatic"
> } 
> ```

I would ignore the "reason" one, as it's currently unused anyway. Meaning ignore whatever is in there.
See what's done currently
```ml
           | Some (Flo { loc; flo }) -> (
               match flo with
               | "3" -> default := Some Jsx_v3
               | _ -> Bsb_exception.errorf ~loc "Unsupported jsx version %s" flo
               )
           | Some x ->
               Bsb_exception.config_error x
                 "Unexpected input (expect a version number) for jsx, note \
                  boolean is no longer allowed"
           | None -> ()) )

```

I would leave that code unchanged.

Btw the schema is here:
https://github.com/rescript-lang/rescript-compiler/blob/master/docs/docson/build-schema.json

mattdamon108:
I've checked that the "reason" is working for the trigger of ppx to rewrite only. Yes we can leave as it is. I'll make a PR for new configuration to the compiler as you introduced. Is it planned to be shiped with v10.1 right?

mattdamon108:
Let me back to the original question. In this PR, do you think we have to put the both ppx v3 and v4 for user to avoid breaking their project?

cristianoc:
> Let me back to the original question. In this PR, do you think we have to put the both ppx v3 and v4 for user to avoid breaking their project?

Definitely yes both v3 and v4.
And a little more likely needed later when thinking about smooth upgrade.

cristianoc:
> I've checked that the "reason" is working for the trigger of ppx to rewrite only. Yes we can leave as it is. I'll make a PR for new configuration "jsx" and build schema to the compiler as you introduced. Is it planned to be shiped with v10.1 right?

10.1 seems like a good target, unless some big unexpected problem comes up -- but at the moment things are going smoothly

Basically, people need to be able to try a pre-release and be able to adopt/upgrade without issues

cristianoc:
How would you like to proceed.
By separating into v3 and v4 it's possible to review and merge this in soon, and keep on refining in a follow-on PR.

mattdamon108:
Sure, why not. I prefer keeping the open time of PR short either. Maybe, we can ship the ppx v4 as an experimental feature with V10 release, it allows us to get feedback and chances to refine the v4 and rescript-react.
As a side comment, I'd like to ask the core team for any chance to update the rescript-react for the official release ppx v4. I think there are missing bindings for the new jsx transform and it is outdated such as`ReasonReact.fragment`.

cristianoc:
How about putting bindings for the new jsx transform directly inside the compiler? Under the `Js.*` namespace.
As from what I understand is not bound to react necessarily.

cli/res_cli.ml
```diff
@@ -203,8 +205,12 @@ end = struct
         "Parse as interface" );
       ( "-ppx",
         Arg.String (fun txt -> ppx := txt),
-        "Apply a specific built-in ppx before parsing, none or jsx. Default: \
-         none" );
+        "Apply a specific built-in ppx before parsing, none or jsx3, jsx4. \
+         Default: none" );
+      ( "-jsx-runtime",
+        Arg.String (fun txt -> jsxRuntime := txt),
+        "Specify the jsx runtime for React, classic or automatic. Default: \
```
cristianoc, line 212: This option only affects V4 right?

src/res_core.ml
```diff
@@ -2604,6 +2605,28 @@ and parseJsxProp p =
           if optional then Asttypes.Optional name else Asttypes.Labelled name
         in
         Some (label, attrExpr))
+  (* {...props} *)
+  | Lbrace -> (
```
cristianoc, line 2609: What happens when this is parsed while V3 is active?
Should this be gated on version or is it OK like this?

cristianoc (commented):
Some minor comments.
Just checking that nothing is broken in V3 support at the moment.
Not really reviewing V4 fully.

cristianoc:
I think what's out there, modulo a couple of pending questions can be merged. And work on V4, and review on it, can continue separately.
So that one does not need to keep on rebasing this PR when master changes.
One obvious thing to review is the use of locations, and what happens with editor integration. But it's OK to do that later.

cli/res_cli.ml
mattdamon108: Yes, it affects V4 only.

src/res_core.ml
mattdamon108: It will be OK with V3. V9 will give us a parsing error, but V10 will parse the JSX application as below, but it will be a type error anyway. Because the definition doesn't have a `spreadProps` as a prop. IMO, it doesn't make anything worse.
```rescript
// lowercase
ReactDOMRe.createDOMElementVariadic(
  "div",
  ~props=ReactDOMRe.domProps(~spreadProps=foo, ()),
  [],
)

// uppercase
React.createElement(Foo.make, Foo.makeProps(~spreadProps=foo, ())),
```

mattdamon108:
> How about putting bindings for the new jsx transform directly inside the compiler? Under the `Js.*` namespace. As from what I understand is not bound to react necessarily.

I can put the new jsx binding into `Js.*`, but can it make an unnecessary dependency of the compiler with react?

cristianoc:
> > How about putting bindings for the new jsx transform directly inside the compiler? Under the `Js.*` namespace. As from what I understand is not bound to react necessarily.
> 
> I can put the new jsx binding into `Js.*`, but can it make an unnecessary dependency of the compiler with react?
There's an entire ppx in the compiler. A couple of bindings don't seem like a big deal.
But, would like to see what they look like / how they are used to know better.

mattdamon108:
> > How about putting bindings for the new jsx transform directly inside the compiler? Under the `Js.*` namespace. As from what I understand is not bound to react necessarily.
> 
> I can put the new jsx binding into `Js.*`, but can it make an unnecessary dependency of the compiler with react?

How about adding some missing binding into rescript-react, and make additional type such as `type domProps2`, `type props` in ReactDOM.res with `@optional` attribute. I think it will not break anything.

cristianoc:
I think this can be merged right now, then continued.

mattdamon108:
Thanks for merging, I'm gonna make the new PR part 2.

PR #548, opened 2022-06-15
ocamlformat

cknitt:
As discussed with @cristianoc, add ocamlformat to the project, but do not reformat files or enforce formatting in CI yet.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,0 +1,11 @@
+profile = default
+version = 0.22.4
+
+field-space = tight-decl
+break-cases = toplevel
+module-item-spacing = preserve
+cases-exp-indent = 2
+space-around-arrays = false
+space-around-lists = false
+space-around-records = false
+space-around-variants = false
--- a/src/res_utf8.ml
+++ b/src/res_utf8.ml
@@ -34,17 +34,16 @@ let locb = 0b1000_0000
 let hicb = 0b1011_1111
 
 let categoryTable = [|
-(* 0 *) {low = -1; high= -1; size= 1}; (* invalid *)
-(* 1 *) {low = 1; high= -1; size= 1}; (* ascii *)
-(* 2 *) {low = locb; high= hicb; size= 2};
-(* 3 *) {low = 0xA0; high= hicb; size= 3};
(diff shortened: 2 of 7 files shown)


.ocamlformat
```diff
@@ -0,0 +1,6 @@
+profile = default
+version = 0.22.4
+
```
cristianoc, line 3: See: https://github.com/rescript-lang/rescript-vscode/blob/master/analysis/.ocamlformat
There's nothing set in stone in that file, just that's been used so far.
Seems like a good idea to converge to one, and use it in all places.

cristianoc:
Looking good.
Just need to make sure there are no big outstanding PRs at merge time.

.ocamlformat
cknitt: I have taken over the additional settings from there now. I am trying to minimize the changes compared to the existing code style.

I am wondering if it is possible to get rid of the space before the `:` in

```ocaml
val loc : t -> Location.t
```

Haven't found the setting for that yet (only for `let` bindings).

src/res_utf8.ml
```diff
@@ -24,118 +24,365 @@ let h4 = 0b1111_0000
...
-  0; 0; 2; 2;2; 2; 2; 2;2; 2; 2; 2;2; 2; 2; 2;
-   2; 2; 2; 2; 2; 2; 2; 2; 2; 2; 2; 2; 2; 2; 2; 2;
-   3; 4; 4; 4; 4; 4; 4; 4; 4; 4; 4; 4; 4; 5; 4; 4;
-  6; 7; 7 ;7; 8; 0; 0; 0; 0; 0; 0; 0; 0; 0; 0; 0;
-|]
+let categories =
+  [|
+    1;
+    1;
+    1;
```
cknitt, line 53: @cristianoc This is not ideal. Probably should put the tables into a separate file that is excluded from formatting.

cknitt: Ah there is an easier way with `[@@ocamlformat "disable"]`.

cknitt:
Now also checking for formatting in CI.

cristianoc (approved):
Great, thanks for doing this!

PR #554, opened 2022-06-16
Improve debug logging

cknitt:


src/res_comments_table.ml
```diff
@@ -75,18 +48,23 @@ let log t =
...
       Doc.line;
-      Doc.indent (Doc.concat leadingStuff);
-      Doc.line;
+      Doc.text "comments inside:";
+      Doc.indent (Doc.concat [Doc.line; Doc.concat stuffInside]);
       Doc.line;
       Doc.text "trailing comments:";
-      Doc.indent (Doc.concat trailingStuff);
-      Doc.line;
+      Doc.indent (Doc.concat [Doc.line; Doc.concat trailingStuff]);
```
cristianoc, line 67: Maybe we should have some simple test that triggers this code path.

PR #556, opened 2022-06-16
Reanalyze failure should not prevent running tests.

cristianoc:
Now running reanalyze only after the normal tests.
Since introducing the exit code, the dev experience has been too strict, where even introducing an unused variable would prevent tests from running.

This also prints all tests passed at the end.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,13 +9,17 @@ bootstrap: build
 bench: build
 	dune exec -- bench
 
-test: reanalyze
+test:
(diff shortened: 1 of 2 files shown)


cknitt (approved):
Another option would be not to run reanalyze as part of make (roundtrip-)test" at all but to invoke it separately from ci.yml.

cristianoc:
That would mean build a tool that runs in tens of millisecond then wait minutes for CI.

PR #563, opened 2022-06-18
Test hack AST.

cristianoc:


cristianoc:
Tested on this: https://github.com/zth/rescript-relay/tree/master/example
seems not to create issues.

PR #564, opened 2022-06-18
Fix comment before Pexp_ifthenelse moving into if branch

### LINKED ISSUE #562: Comment jumps inside if expression
```rescript
{ // comment
  if condition() {
    renderBranch()
  } else {
    doSomeOtherThings()
  }
}
```

is printed as

```rescript
{
  if (
    // comment
    condition()
  ) {
    renderBranch()
  } else {
    doSomeOtherThings()
  }
}
```

cknitt:
This fixes #562.

### DIFF
--- a/src/res_comments_table.ml
+++ b/src/res_comments_table.ml
@@ -1088,7 +1088,9 @@ and walkExpression expr t comments =
       walkExpression expr2 t inside;
       attach t.trailing expr2.pexp_loc trailing
   | Pexp_ifthenelse (ifExpr, thenExpr, elseExpr) -> (
-    let leading, inside, trailing = partitionByLoc comments ifExpr.pexp_loc in
+    let leading, rest = partitionLeadingTrailing comments expr.pexp_loc in
+    attach t.leading expr.pexp_loc leading;
+    let leading, inside, trailing = partitionByLoc rest ifExpr.pexp_loc in
     let comments =
       if isBlockExpr ifExpr then (
         let afterExpr, comments =
--- a/tests/printer/comments/expected/expr.res.txt
+++ b/tests/printer/comments/expected/expr.res.txt
@@ -140,10 +140,8 @@ if /* c0 */ user.name === "John" /* c1 */ {
 
 if a {
   b
-} else if (
-  // test
-  c
-) {
+} // test
+else if c {
   d
 }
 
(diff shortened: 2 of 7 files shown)


src/res_comments_table.ml
```diff
@@ -1088,7 +1088,9 @@ and walkExpression expr t comments =
       walkExpression expr2 t inside;
       attach t.trailing expr2.pexp_loc trailing
   | Pexp_ifthenelse (ifExpr, thenExpr, elseExpr) -> (
-    let leading, inside, trailing = partitionByLoc comments ifExpr.pexp_loc in
+    let leading, rest = partitionLeadingTrailing comments expr.pexp_loc in
+    attach t.leading expr.pexp_loc leading;
+    let leading, inside, trailing = partitionByLoc rest ifExpr.pexp_loc in
```
cristianoc, line 1093: This logic makes perfect sense.
Question: why is this not the logic applied to every node?
(except value and type def for example)

cristianoc (approved):

src/res_comments_table.ml
cristianoc: E.g. isn't `Pexp_setfield` suffering from the same issue?

cknitt: I don't know really - do you mean doing this at the beginning of `walkExpression` before the match on `pexp_desc`?
Maybe @IwanKaramazow can comment.

cristianoc: Don't even know if this is only for expressions.

tests/printer/comments/expected/expr.res.txt
```diff
@@ -155,10 +153,8 @@ let distXToTraverse = if deltaX < 0. {
   } else {
     deltaX +. minRangeX
   }
-} else if (
-  // wizard to the right
-  minRangeX >= deltaX
-) {
+} // wizard to the right
```
IwanKaramazow, line 156: The printed code doesn't look that "smooth".

Can you log a separate issue to check if we can preserve nested if else branches?
```
if a {
  b
} else {
  // test
  if c {
    d
  }
}

```

src/res_comments_table.ml
IwanKaramazow: I can't remember, the comment table mostly evolved in an ad-hoc way… It does make sense to apply this logic to every node.

tests/printer/comments/expected/expr.res.txt
cknitt: I'll log an issue for tracking, but I actually thought the flattening was a feature here. Of course, if flattening, it is a bit of an edge case that a comment is between the else and the if. The user can still move the comment down manually after formatting in this case.

IwanKaramazow: Maybe it's only a feature when we don't have comments? we can think about it later

Issue #429, opened 2021-05-29
Escaping a backslash in backtick strings doesn't work well


joakin:
See https://forum.rescript-lang.org/t/strings-and-how-to-escape-characters-in-the-different-types-of-string-literals/1759/1 for context.

When trying to build a string that when logged will output `\a -> a`, with double quotes the escaping of the backslash works like in JS, but with the backtick string there is a strange error, and when using 3 backslashes it does work but it may be a coincidence.

```
let f = "\\a -> a" // Fine, as expected

// Doesn't compile:
// [E] Line 6, column 9:
// Offset: 1, Invalid escape code: a
let f' = `\\a -> a`

let f'' = `\\\a -> a` // Triple backslash makes it work for some reason?
```
[Playground link](https://rescript-lang.org/try?code=DYUwLgBAZhC8ECIA6SCGEC0A+CqECh8B6IiAEQHsQBnAOwHJIBjCgWwAcBLUALmNIDaAUQC6EADKdaICADYANBBbAArq1oQAnHxIQA8lCjVwPCAEZFASVoA3VME4ATCDSap2Mlo5CnU-CKCQUPRwEAAGKOjYuGGEgdD0IfARkZg4qLFAA)

IwanKaramazow:
Thanks for the bug report, I will have a look.

johnridesabike:
Is this related to #161?

IwanKaramazow:
Yes, I'm looking at the escaping rules.

PR #569, opened 2022-06-18
Don't escape template literal string, just as normal strings are not …

cristianoc:
…escaped.

Fixes https://github.com/rescript-lang/syntax/issues/429


src/res_core.ml
```diff
@@ -1020,10 +1020,6 @@ let parseTemplateConstant ~prefix (p : Parser.t) =
   match p.token with
   | TemplateTail txt ->
     Parser.next p;
-    let txt =
-      if p.mode = ParseForTypeChecker then parseTemplateStringLiteral txt
```
cristianoc, line 1024: One difference with normal strings is that normal strings do:
```ml
      if p.mode = ParseForTypeChecker then Pconst_string (s, Some "js")
      else Pconst_string (s, None)
```
but I guess here we don't want to change the prefix.

CC @IwanKaramazow

cristianoc: This is left in case of `#foo`:

```ml
            let text =
              if p.mode = ParseForTypeChecker then parseStringLiteral text
              else text
            in
```
and several other cases. Is that intentional

Also, there's the case of `TemplatePart` that needs looking into, and is not changed yet.

IwanKaramazow: Right, I guess for poly-variants with string literals: `#"string here"`. The question is if what happens when we have unicode and/or escape codes in here.

cristianoc: > Right, I guess for poly-variants with string literals: `#"string here"`. The question is if what happens when we have unicode and/or escape codes in here.

Does the compiler care if there's unicode in the name of a poly variant? I guess it does not look at the string at all.
Is the question about when printing a type error?

IwanKaramazow: This might be outside of the scope of this PR, but the question is what we want here:
image

IwanKaramazow (approved):

PR #574, opened 2022-06-22
Put each jsx tag on own line

### LINKED ISSUE #570: Put each JSX tag on its own line
Currently, when printing JSX, the printer tries to fit as much as possible on one line.

As a contrived example,

```rescript
<A>
  <B>
    <C>
      <D />
      <E />
    </C>
    <F>
      <G />
      <H />
    </F>
  </B>
</A>
```

will be printed as

```rescript
<A> <B> <C> <D /> <E /> </C> <F> <G /> <H /> </F> </B> </A>
```

which is rather unreadable.

Looking at other code formatters, [Prettier](https://prettier.io/) actually does the reverse, it expands the one liner so that each JSX tag is on its own line. We should change the ReScript printer to do the same.

nkrkv:
Closes #570 

This is my first lines of code in OCaml, so I’m not sure I’m doing it right :smiley_cat:

### DIFF
--- a/src/res_printer.ml
+++ b/src/res_printer.ml
@@ -3694,6 +3694,13 @@ and printJsxExpression lident args cmtTbl =
       true
     | _ -> false
   in
+  let hasNestedJsx =
+    match children with
+    | Some expr ->
+      let childExprs, _ = ParsetreeViewer.collectListExpressions expr in
+      childExprs |> List.exists ParsetreeViewer.isJsxExpression
+    | None -> false
+  in
   Doc.group
     (Doc.concat
        [
@@ -3728,10 +3735,12 @@ and printJsxExpression lident args cmtTbl =
                       Doc.line;
                       (match children with
                       | Some childrenExpression ->
-                        printJsxChildren childrenExpression cmtTbl
+                        printJsxChildren childrenExpression
+                          ~sep:(if hasNestedJsx then Doc.hardLine else Doc.line)
+                          cmtTbl
                       | None -> Doc.nil);
                     ]);
-               Doc.line;
+               (if hasNestedJsx then Doc.hardLine else Doc.line);
                Doc.text "</";
                name;
                Doc.greaterThan;
@@ -3748,17 +3757,19 @@ and printJsxFragment expr cmtTbl =
          opening;
          (match expr.pexp_desc with
          | Pexp_construct ({txt = Longident.Lident "[]"}, None) -> Doc.nil
-         | _ -> Doc.indent (Doc.concat [Doc.line; printJsxChildren expr cmtTbl]));
+         | _ ->
+           Doc.indent
+             (Doc.concat [Doc.line; printJsxChildren expr ~sep:Doc.line cmtTbl]));
          Doc.line;
          closing;
        ])
 
-and printJsxChildren (childrenExpr : Parsetree.expression) cmtTbl =
+and printJsxChildren (childrenExpr : Parsetree.expression) ~sep cmtTbl =
   match childrenExpr.pexp_desc with
   | Pexp_construct ({txt = Longident.Lident "::"}, _) ->
     let children, _ = ParsetreeViewer.collectListExpressions childrenExpr in
     Doc.group
-      (Doc.join ~sep:Doc.line
+      (Doc.join ~sep
          (List.map
             (fun (expr : Parsetree.expression) ->
               let leadingLineCommentPresent =
--- a/src/res_printer.ml
+++ b/src/res_printer.ml
@@ -3750,7 +3750,8 @@ and printJsxExpression lident args cmtTbl =
 and printJsxFragment expr cmtTbl =
   let opening = Doc.text "<>" in
   let closing = Doc.text "</>" in
-  (* let (children, _) = ParsetreeViewer.collectListExpressions expr in *)
+  let children, _ = ParsetreeViewer.collectListExpressions expr in
+  let hasNestedJsx = children |> List.exists ParsetreeViewer.isJsxExpression in
   Doc.group
     (Doc.concat
        [
@@ -3759,8 +3760,14 @@ and printJsxFragment expr cmtTbl =
          | Pexp_construct ({txt = Longident.Lident "[]"}, None) -> Doc.nil
          | _ ->
            Doc.indent
-             (Doc.concat [Doc.line; printJsxChildren expr ~sep:Doc.line cmtTbl]));
-         Doc.line;
+             (Doc.concat
+                [
+                  Doc.line;
+                  printJsxChildren expr
+                    ~sep:(if hasNestedJsx then Doc.hardLine else Doc.line)
+                    cmtTbl;
+                ]));
+         (if hasNestedJsx then Doc.hardLine else Doc.line);
          closing;
        ])
 
--- a/src/res_printer.ml
+++ b/src/res_printer.ml
@@ -99,12 +99,12 @@ let hasCommentBelow tbl loc =
   | [] -> false
   | exception Not_found -> false
 
-let hasNestedJsx expr =
+let rec hasNestedJsx expr =
   match expr.Parsetree.pexp_desc with
   | Pexp_construct
-      ({txt = Longident.Lident "::"}, Some {pexp_desc = Pexp_tuple (hd :: [_])})
-    ->
-    ParsetreeViewer.isJsxExpression hd
+      ( {txt = Longident.Lident "::"},
+        Some {pexp_desc = Pexp_tuple (hd :: [tail])} ) ->
+    ParsetreeViewer.isJsxExpression hd || hasNestedJsx tail
   | _ -> false
 
 let printMultilineCommentContent txt =
--- a/src/res_printer.ml
+++ b/src/res_printer.ml
@@ -99,13 +99,22 @@ let hasCommentBelow tbl loc =
   | [] -> false
   | exception Not_found -> false
 
-let rec hasNestedJsx expr =
-  match expr.Parsetree.pexp_desc with
-  | Pexp_construct
-      ({txt = Longident.Lident "::"}, Some {pexp_desc = Pexp_tuple [hd; tail]})
-    ->
-    ParsetreeViewer.isJsxExpression hd || hasNestedJsx tail
-  | _ -> false
+let hasNestedJsxOrManyChildren expr =
+  let hasConent pexp_desc =
+    match pexp_desc with
+    | Parsetree.Pexp_construct ({txt = Longident.Lident "[]"}, _) -> false
+    | _ -> true
+  in
+  let rec loop inRecursion expr =
+    match expr.Parsetree.pexp_desc with
+    | Pexp_construct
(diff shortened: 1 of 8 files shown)


cknitt:
This is awesome! Thanks a lot!

Just wondering about performance when computing `hasNestedJsx` -  I think we could avoid the list creation done by `ParsetreeViewer.collectListExpressions` here.

/cc @IwanKaramazow

cristianoc:
> This is awesome! Thanks a lot!
> 
> Just wondering about performance when computing `hasNestedJsx` - I think we could avoid the list creation done by `ParsetreeViewer.collectListExpressions` here.
> 
> /cc @IwanKaramazow

Indeed.
Bonus point: do it with without allocating.
For example, creating a list and walking a list is allocating.
Technically, even creating a single closure to pass to `List.exists` would be allocating, but that's a very small allocation.

nkrkv:
> I think we could avoid the list creation done by `ParsetreeViewer.collectListExpressions` here.
> Bonus point: do it with without allocating.

Thank you gentlemen. Would you point me a little bit, how can I traverse expression children without creating a list? Just a function name, a very rough idea, or a hint is enough. Otherwise I can go wrong direction again :)

cknitt:
I'd say iterate/recurse like `ParsetreeViewer.collectListExpressions` does, but do not build a list, instead directly check each item if its a JSX expression (`ParsetreeViewer.isJsxExpression`). Return true if a JSX expression was found, otherwise continue recursion.

cristianoc:
> I'd say iterate/recurse like `ParsetreeViewer.collectListExpressions` does, but do not build a list, instead directly check each item if its a JSX expression (`ParsetreeViewer.isJsxExpression`). Return true if a JSX expression was found, otherwise continue recursion.

So create a little helper function which is a recursive traversal.
Then call that function.
This will also take care of the little code duplication in the current PR.

src/res_printer.ml
```diff
@@ -99,6 +99,15 @@ let hasCommentBelow tbl loc =
   | [] -> false
   | exception Not_found -> false
 
+let rec hasNestedJsx expr =
+  match expr.Parsetree.pexp_desc with
+  | Pexp_construct ({txt = Longident.Lident "[]"}, _) -> false
+  | Pexp_construct
+      ( {txt = Longident.Lident "::"},
+        Some {pexp_desc = Pexp_tuple (hd :: [tail])} ) ->
```
cristianoc, line 107: nit: this is normally written `[hd, tail]`.

src/res_printer.ml
```diff
@@ -99,6 +99,15 @@ let hasCommentBelow tbl loc =
...
+    hasNestedJsx hd || hasNestedJsx tail
```
cristianoc, line 108: Should this be `ParsetreeViewer.isJsxExpression hd` ? and the next line do nothing? (and just return false)

cristianoc (approved):
Looks great.
Some tiny style comments.

cristianoc:
This is looking great.
As a final step, before and after: `make bench`. See what it says about allocation.

cristianoc:
If you are on `macos`, that is. Does not work on other architectures at the moment.

src/res_printer.ml
cknitt: Yes. Otherwise it behaves differently than before this commit (see the spread change in this commit which I think we do not want).

src/res_printer.ml
nkrkv: Hmmm, this gives me:

```
File "src/res_printer.ml", line 106, characters 39-44:
106 |         Some {pexp_desc = Pexp_tuple ([hd, _])} ) ->
                                             ^^^^^
Error: This pattern matches values of type 'a * 'b
       but a pattern was expected which matches values of type
         Parsetree.expression
```

My OCaml knowledge is not enough to understand why is it different. Perhaps, type inference fails somehow it the `[hd, tail]` case. Anyway, I’ve stolen this pattern from `ParsetreeViewer.collectListExpressions`.

nkrkv:
> As a final step, before and after: make bench ... If you are on macos, that is

I’m on Linux and have no access to a macOS machine, unfortunately. Would someone help with running `bench`?

src/res_printer.ml
cristianoc: Sorry `[hd; tail]`

cristianoc: > > As a final step, before and after: make bench ... If you are on macos, that is
> 
> I’m on Linux and have no access to a macOS machine, unfortunately. Would someone help with running `bench`?

Will do once the PR is final.

src/res_printer.ml
```diff
@@ -99,6 +99,14 @@ let hasCommentBelow tbl loc =
   | [] -> false
   | exception Not_found -> false
 
+let hasNestedJsx expr =
+  match expr.Parsetree.pexp_desc with
+  | Pexp_construct
+      ({txt = Longident.Lident "::"}, Some {pexp_desc = Pexp_tuple (hd :: [_])})
+    ->
+    ParsetreeViewer.isJsxExpression hd
```
cristianoc, line 107: What about some children are, and some are not, JSX expressions?

cristianoc: Perhaps a good test to add.

cknitt: Yes, I don't think this is correct without recursion. If hd is not a JSX expression, we still need to recurse over tail.

E.g. for cases like

```rescript
<A>
  {children}
  <B/>
<A/>
```

cknitt:
Hmm, and what do we actually do if we have a lot of children, but none is a JSX expression?

Like
```rescript
<A> {a} {b} {c} {d} {e} </A>
```

Maybe the logic would need to be something like: expand if a child is a JSX expression or if there is more than 1 child?

What do you think @cristianoc?

cristianoc:
> Hmm, and what do we actually do if we have a lot of children, but none is a JSX expression?
> 
> 
> 
> Like
> 
> ```rescript
> 
> <A> {a} {b} {c} {d} {e} </A>
> 
> ```
> 
> 
> 
> Maybe the logic would need to be something like: expand if a child is a JSX expression or if there is more than 1 child?
> 
> 
> 
> What do you think @cristianoc?

What does prettier do?

src/res_printer.ml
nkrkv: Ouch, sure. I’ve recovered the recursion add added the tests

src/res_printer.ml
nkrkv: > Sorry `[hd; tail]` not `[hd, tail]`

Muscle memory :) Fixed it.

nkrkv:
> What does prettier do?

```jsx
<A>{a} {b} {c} {d}</A>;
```

:point_down: 

```jsx
<A>
  {a} {b} {c} {d}
</A>;
```

Do we want the same? It would made the algorithm slightly more complex (line separator between children and after/before tags differs) but doable nevertheless.

cknitt:
I tried 

```js
<A>{a}{b}{c}{d}</A>;
```

in one of our JS projects using prettier and got

```js
<A>
  {a}
  {b}
  {c}
  {d}
</A>;
```

Maybe our prettier config is different than yours.

cknitt:
And

```js
<A>{a}</A>;
```

stays unmodified for me.

nkrkv:
It is from their [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAggPmAQwL4AEwARocGGQCZlx6oD0WIANCBAA4wCW0AzsqBwAnIRADuABWEJ+KHABsxOAJ79WxITjABrODADKOALZwAMlyhxkAMwW846zTr372WiwHNkMIQFcHIHBGxHCUlKGmOFAevjgecABiEEJGODDc0cggOL4wECwgABYwRvIA6oVc8LxuYHD6MlVcAG5VyllgvGogFvZCMBKaHqk2dgEAVrwAHvqe8nAAir4Q8KPy9qxuQn1ZxDgh8gXsQhYwZVyUMIXIABwADJui9mWa7FnHcH3NVqwAjsvwQYcWTZXgAWksoVCBSEcH+XFhgziIyQtnWAXsRi43j8GLmiwBVlRY1YMH250u1yQACZSZouPJPABhCBGFGBXgAVgKvnsABV9rI0RsQM1-ABJKDhWD6MAnTjoKX6GDKeZrex4PBAA) with default settings

cknitt:
> Maybe the logic would need to be something like: expand if a child is a JSX expression or if there is more than 1 child?

actually matches Prettier's behavior as far as I can see:

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAggPlQIQPTZ7oAEmA3ADpRUbb7YDCBqRpl1UtuBwYAvixLkqNLMABGApoSHtRmYAEM+Ega2EdaqwWyogANCAgAHGAEtoAZ2ShFAJzsQA7gAV7CaykUAbJ4oCe1obidopgANZwMADKigC2cAAyZlBwyABmPpZwwaERUdHGYSkA5sgwdgCuOSBwceJwACaNTYmKUCWViiVwAGIQdnGKMOYdyCCKlTAQBiAAFjBx3gDqc2bwlkVgcNEe62YAbuv+42CWQSAp2XYwLqElQxlZNQBWlgAe0aXecACKlRB4E9vNlDEU7NdxuJFA1vLNjHYUjBlmZGjA5sgABwABjBjmyy1CxnGCLg1wOaUMAEcAfA7iZPBNLABaVJNJqzOxwGlmLl3bqPJCZEE1bJxMzlKqi75-WlpIXPQwwGEotEYpAAJiVoTM3lKDAgcUFtUsAFZZpVsgAVGGeYWgkAHaoASSgLVg0TAiNM6Dd0Rg-h+wOyfD4QA

tests/printer/other/expected/fatSlider.res.txt
```diff
@@ -50,7 +50,8 @@ let make = (~min=50, ~max=250, ~meterSuffix=?) => {
       />
     </div>
     <div className="meter text-4xl font-bold text-gray-600 mt-4">
-      {values[0]->Js.Int.toString->string} {meterSuffix->Belt.Option.getWithDefault(null)}
+      {values[0]->Js.Int.toString->string}
+      {meterSuffix->Belt.Option.getWithDefault(null)}
```
cknitt, line 54: Definitely an improvement in this example!

cknitt:
I think it can be done in one match without the "isConent":

```rescript
let hasNestedJsxOrMoreThanOneChild expr =
  let rec loop inRecursion expr =
    match expr.Parsetree.pexp_desc with
    | Pexp_construct ({txt = Longident.Lident "[]"}, _) -> false
    | Pexp_construct
        ({txt = Longident.Lident "::"}, Some {pexp_desc = Pexp_tuple [hd; tail]})
      ->
      if inRecursion || ParsetreeViewer.isJsxExpression hd then true
      else loop true tail
    | _ -> false
  in
  loop false expr
```

nkrkv:
Yes, thanks. This is definitely more clear and I like the name improvement.

cknitt (approved):
Thanks! 👍

PR #575, opened 2022-06-23
Js strict mode

cristianoc:
Continues https://github.com/rescript-lang/syntax/pull/470


src/res_string.ml
```diff
@@ -1,4 +1,22 @@
...
+    '7';
+    '8';
+    '9';
+    'a';
+    'b';
+    'c';
+    'd';
+    'e';
+    'f';
+  |]
```
cknitt, line 19: If you like to keep this as a one-liner, you can use `[@@ocamlformat "disable"]` here.

cristianoc:
This is currently a breaking change as it makes some `.res` code not parse anymore.
Plus, a bunch of `.ml` roundtrip tests cannot pass as they are not representable in ReScript.

So the underlying issue is that the generated code is not strict JS.
One way to go about it is to emit valid strict JS, and do not introduce breaking changes.
So:
- Even when the source containts `let dd = "\123"` parse successfully and emit strict JS.
- Reformat it so that the source itself uses strict JS, i.e. to `let dd = "\x53"`.

cristianoc:
Roundtrip tests are not going to succeed.
As the OCaml parser will put `\123` in the AST, but the next step when parser from a `.res` file it will be stored in the AST as `x7b`.

IwanKaramazow:
Since we got rid of Reason, maybe we can also drop the ml round trip?

cristianoc:
Removing the .ml roundtrip tests brings down the time by about 30%.

Issue #261, opened 2021-02-12
Formatter hangs on deeply nested function calls


nkrkv:
### Steps to reproduce

Put the following in a file (say, `Hang.res`):

```rescript
let foo = () =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x =>
  bar(x => x)
  ))))))))))))))
```

Run:

```bash
bsc -format Hang.res
```

## Expected result

The source is indented properly and printed to stdout

## Actual result

The process hangs, loads CPU to 100%, eats one GB of RAM after another, then after 1m30s or so crashes with out-of-memory.

The source (if put back into the context), _compiles_ fine. Just formatting causes the problem.

## Machine info

```console
$ node --version                                               
v12.16.3

$ bsc -version        
BuckleScript 9.0.0 ( Using OCaml:4.06.1+BS )

$ uname -a            
Linux archxx 5.10.6-arch1-1 #1 SMP PREEMPT Sat, 09 Jan 2021 18:22:35 +0000 x86_64 GNU/Linux
```

IwanKaramazow:
Thanks for the report. I will investigate this over the weekend. This code is probably generated by a ppx?

cristianoc:
Part involved: https://github.com/rescript-lang/syntax/blob/master/src/res_printer.ml#L4059

nkrkv:
>  This code is probably generated by a ppx?

Nope, it’s hand-written (simplified version) to convert 15-tuple of `Result.t` to `Result.t` of 15-tuple

bobzhang:
might be there is some exponential complexity iN the algorithm

cristianoc:
We already found the issue some time back.
It is an exponential blowup.
There are 3 possible formatting s for callbacks, and the first one that fits line width is chosen.

With n nested callbacks it's 3^n cases. The cases are generated statically, and the explosion happens at case generation time.

One needs to be a little creative here. By Eg prioritize one case when there are nested callbacks. Or use some other creative solution.

leeor:
Started running into this issue more now that I'm removing let-syntax from our code-base. It hits tests files pretty hard because they always involve at least two callbacks before any "real" code (describe + test boilerplate). Some of which I have to break into multiple files just because they can't be formatted.

[...]

PR #576, opened 2022-06-23
improve printing algorithm to performance for many nested callbacks

leeor:
Suggested implementation, with a caveat of me know being familiar with the code at all :-)

I have completely removed `propagateForcedBreaks` from the code as it looked to me to be doing nothing while forcing evaluation of everything. Did I miss something?

See #261.

```
time ./_build/install/default/bin/rescript -print res <test file>
./_build/install/default/bin/rescript -print res   0.65s user 0.03s system 33% cpu 2.009 total
```

### DIFF
--- a/src/res_doc.ml
+++ b/src/res_doc.ml
@@ -21,7 +21,7 @@ type t =
   | LineSuffix of t
   | LineBreak of lineStyle
   | Group of {mutable shouldBreak: bool; doc: t}
-  | CustomLayout of t list
+  | CustomLayout of (t Lazy.t) list
   | BreakParent
 
 let nil = Nil
@@ -72,54 +72,13 @@ let equal = Text "="
 let trailingComma = ifBreaks comma nil
 let doubleQuote = Text "\""
 
-let propagateForcedBreaks doc =
-  let rec walk doc =
-    match doc with
-    | Text _ | Nil | LineSuffix _ -> false
-    | BreakParent -> true
-    | LineBreak (Hard | Literal) -> true
-    | LineBreak (Classic | Soft) -> false
-    | Indent children ->
-      let childForcesBreak = walk children in
-      childForcesBreak
-    | IfBreaks ({yes = trueDoc; no = falseDoc} as ib) ->
-      let falseForceBreak = walk falseDoc in
-      if falseForceBreak then (
-        let _ = walk trueDoc in
-        ib.broken <- true;
-        true)
-      else
-        let forceBreak = walk trueDoc in
-        forceBreak
-    | Group ({shouldBreak = forceBreak; doc = children} as gr) ->
-      let childForcesBreak = walk children in
-      let shouldBreak = forceBreak || childForcesBreak in
-      gr.shouldBreak <- shouldBreak;
-      shouldBreak
-    | Concat children ->
-      List.fold_left
-        (fun forceBreak child ->
-          let childForcesBreak = walk child in
-          forceBreak || childForcesBreak)
-        false children
-    | CustomLayout children ->
-      (* When using CustomLayout, we don't want to propagate forced breaks
-       * from the children up. By definition it picks the first layout that fits
-       * otherwise it takes the last of the list.
-       * However we do want to propagate forced breaks in the sublayouts. They
-       * might need to be broken. We just don't propagate them any higher here *)
-      let _ = walk (Concat children) in
-      false
-  in
-  let _ = walk doc in
-  ()
-
 (* See documentation in interface file *)
 let rec willBreak doc =
   match doc with
   | LineBreak (Hard | Literal) | BreakParent | Group {shouldBreak = true} ->
     true
-  | Group {doc} | Indent doc | CustomLayout (doc :: _) -> willBreak doc
+  | Group {doc} | Indent doc -> willBreak doc
+  | CustomLayout (doc :: _) -> willBreak (Lazy.force doc)
   | Concat docs -> List.exists willBreak docs
   | IfBreaks {yes; no} -> willBreak yes || willBreak no
   | _ -> false
@@ -157,7 +116,7 @@ let fits w stack =
     | _, Concat docs -> calculateConcat indent mode docs
     | _, CustomLayout (hd :: _) ->
       (* TODO: if we have nested custom layouts, what we should do here? *)
-      calculate indent mode hd
+      calculate indent mode (Lazy.force hd)
     | _, CustomLayout [] -> ()
   and calculateConcat indent mode docs =
     if result.contents == None then
@@ -178,7 +137,6 @@ let fits w stack =
   calculateAll stack
 
 let toString ~width doc =
-  propagateForcedBreaks doc;
   let buffer = MiniBuffer.create 1000 in
 
   let rec process ~pos lineSuffices stack =
@@ -238,9 +196,10 @@ let toString ~width doc =
         let rec findGroupThatFits groups =
           match groups with
           | [] -> Nil
-          | [lastGroup] -> lastGroup
+          | [lastGroup] -> Lazy.force lastGroup
           | doc :: docs ->
-            if fits (width - pos) ((ind, Flat, doc) :: rest) then doc
+            let computedDoc = Lazy.force doc in
+            if fits (width - pos) ((ind, Flat, computedDoc) :: rest) then computedDoc
             else findGroupThatFits docs
         in
         let doc = findGroupThatFits docs in
@@ -291,7 +250,7 @@ let debug t =
                (concat
                   [
                     line;
-                    join ~sep:(concat [text ","; line]) (List.map toDoc docs);
+                    join ~sep:(concat [text ","; line]) (List.map (fun doc -> toDoc (Lazy.force doc)) docs);
                   ]);
              line;
              text ")";
--- a/src/res_doc.mli
+++ b/src/res_doc.mli
@@ -16,7 +16,7 @@ val breakableGroup : forceBreak:bool -> t -> t
 (* `customLayout docs` will pick the layout that fits from `docs`.
  * This is a very expensive computation as every layout from the list
  * will be checked until one fits. *)
-val customLayout : t list -> t
+val customLayout : (t Lazy.t) list -> t
 val breakParent : t
 val join : sep:t -> t list -> t
 
--- a/src/res_printer.ml
+++ b/src/res_printer.ml
@@ -1883,12 +1883,12 @@ and printValueBinding ~recFlag vb cmtTbl i =
     if ParsetreeViewer.isSinglePipeExpr vb.pvb_expr then
       Doc.customLayout
         [
-          Doc.group
+          lazy (Doc.group
             (Doc.concat
                [
                  attrs; header; patternDoc; Doc.text " ="; Doc.space; printedExpr;
-               ]);
-          Doc.group
+           ]));
+          lazy (Doc.group
             (Doc.concat
                [
                  attrs;
@@ -1896,7 +1896,7 @@ and printValueBinding ~recFlag vb cmtTbl i =
                  patternDoc;
                  Doc.text " =";
                  Doc.indent (Doc.concat [Doc.line; printedExpr]);
-               ]);
+                 ]));
         ]
     else
       let shouldIndent =
@@ -3936,7 +3936,7 @@ and printArgumentsWithCallbackInFirstPosition ~uncurried args cmtTbl =
    * }, longArgumet, veryLooooongArgument)
    *)
   let fitsOnOneLine =
-    Doc.concat
+    lazy (Doc.concat
       [
         (if uncurried then Doc.text "(. " else Doc.lparen);
         callback;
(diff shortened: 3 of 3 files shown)


leeor:
Removing the `propagate...` function did break a bunch of things :-)

leeor:
The problem is that the way `propagateForceBreaks` is implemented forces evaluation of all elements of the document, the handling of the `Concat` case is a `fold_left` so no early exit condition. It is called from `toString` before the `findGroupThatFits` are evaluated, so it processes elements that will be later discarded.

Can the two cases that `propagate...` handles be handled in the recursive `process` function within `toString`?

cristianoc:
Prettier:
```ts
let foo = function (x, y, z) {
  foo(x, y, z);
};

let foo = function (x, y, z) {
  foo(x, y, function (x, y, z) {
    foo(x, y, z);
  });
};

let foo = function (x, y, z) {
  foo(x, y, function (x, y, z) {
    foo(x, y, function (x, y, z) {
      foo(x, y, z);
    });
  });
};

let foo = function (x, y, z) {
  foo(x, y, function (x, y, z) {
    foo(x, y, function (x, y, z) {
      foo(x, y, function (x, y, z) {
        foo(x, y, z);
      });
    });
  });
};

let foo1 = (x, y, z) => foo(x, y, z);

let foo2 = (x, y, z) => foo1(x, y, (x, y, z) => foo(x, y, z));

let foo3 = (x, y, z) =>
  foo2(x, y, (x, y, z) => foo1(x, y, (x, y, z) => foo(x, y, z)));

let foo4 = (x, y, z) =>
  foo3(x, y, (x, y, z) =>
    foo2(x, y, (x, y, z) => foo1(x, y, (x, y, z) => foo(x, y, z)))
  );

let foo5 = (x, y, z) =>
  foo4(x, y, (x, y, z) =>
    foo3(x, y, (x, y, z) =>
      foo2(x, y, (x, y, z) => foo1(x, y, (x, y, z) => foo(x, y, z)))
    )
  );

let foo1 = (x, y, z) => {
  foo(x, y, z);
};

let foo2 = (x, y, z) => {
  foo1(x, y, (x, y, z) => {
    foo(x, y, z);
  });
};

let foo3 = (x, y, z) => {
  foo2(x, y, (x, y, z) => {
    foo1(x, y, (x, y, z) => {
      foo(x, y, z);
    });
  });
};

let foo4 = (x, y, z) => {
  foo3(x, y, (x, y, z) => {
    foo2(x, y, (x, y, z) => {
      foo1(x, y, (x, y, z) => {
        foo(x, y, z);
      });
    });
  });
};
```


Rescript:
```rescript
let foo1 = (x, y, z) => foo(x, y, z)

let foo2 = (x, y, z) => foo1(x, y, (x, y, z) => foo(x, y, z))

let foo3 = (x, y, z) => foo2(x, y, (x, y, z) => foo1(x, y, (x, y, z) => foo(x, y, z)))

let foo4 = (x, y, z) =>
  foo3(x, y, (x, y, z) => foo2(x, y, (x, y, z) => foo1(x, y, (x, y, z) => foo(x, y, z))))

let foo5 = (x, y, z) =>
  foo4(x, y, (x, y, z) =>
    foo3(x, y, (x, y, z) => foo2(x, y, (x, y, z) => foo1(x, y, (x, y, z) => foo(x, y, z))))
  )

let foo1 = (x, y, z) => {
  foo(x, y, z)
}

let foo2 = (x, y, z) => {
  foo1(x, y, (x, y, z) => {
    foo(x, y, z)
  })
}

let foo3 = (x, y, z) => {
  foo2(x, y, (x, y, z) => {
    foo1(x, y, (x, y, z) => {
      foo(x, y, z)
    })
  })
}

let foo4 = (x, y, z) => {
  foo3(x, y, (x, y, z) => {
    foo2(x, y, (x, y, z) => {
      foo1(x, y, (x, y, z) => {
        foo(x, y, z)
      })
    })
  })
}
```

cristianoc:
On the example


Prettier:
```ts
let foo = () =>
  bar((x) =>
    bar((x) =>
      bar((x) =>
        bar((x) =>
          bar((x) =>
            bar((x) =>
              bar((x) =>
                bar((x) => bar((x) => bar((x) => bar((x) => bar((x) => x)))))
              )
            )
          )
        )
      )
    )
  );
```

ReScript:
```rescript
let foo = () =>
  bar(x =>
    bar(x =>
      bar(x =>
        bar(x => bar(x => bar(x => bar(x => bar(x => bar(x => bar(x => bar(x => bar(x => x)))))))))
      )
    )
  )
```

cristianoc:
So first observation is: prettier might be using 2 things not 3? Possible.

Second observation: the state exploration does not seem to need to be complete (i.e. 2 or 3, to the power of n). In the following sense:
- If the nested one further down breaks, the current one breaks too.

So you don't get e.g. a stack of (breaks, doesn't, breaks, doesn't).
So really it should not be exponential explosion.

In that sense, the semantics of `Doc.customLayout` would change in that it would not try all the possibilities.

How does that sound?

leeor:
Sounds good to me. Is that a complex change?

cristianoc:
> Sounds good to me. Is that a complex change?

Don't know, one needs to try.
I think the printing functions will likely need to pass an extra parameter to know whether it's an immediately-nested case.
But to test the idea, one can just set a global reference in the file. And add a coupe of lines that mutate it.

cristianoc:
I think likely only the `willBreak` and `fits` functions in `res_doc. ml`.

cristianoc:
This is what I would try.

Permitted sequences have the form 1 1 1 1 1 2 2 2 ~1~.

Suppose there are at most 2 cases. Keep the current `level`, a number between 1 and 2.

If the current level is `1` then do the following:
- Take `case1` and try to format the rest at level `1`. If it fits, choose `case1`.
- If it does not fit, choose `case2` and format the rest at level `2`.

If the current level is `2` then:
- Choose `case2` and format the rest at level `2`.

This generalises to level `n` by always trying case `n` and switch to case `n+1` and level `n+1` when it does not fit.

There can be different instances of `customLayout` with different cases, and different numbers of cases. One can tree them all the same. Start from level 1, and if the current level is, say 4, and the `customLayout` being considered only has 2 cases,  
assume by convention that we're talking about `case2`.

leeor:
How does your suggestion solve the performance hit incurred by scanning the whole tree in `propagateForceBreaks` immediately when `toString` starts?

Maybe `toString` can be refactored to incorporate that logic but only for the actual nodes that are chosen?

cristianoc:
> How does your suggestion solve the performance hit incurred by scanning the whole tree in `propagateForceBreaks` immediately when `toString` starts?
> 
> Maybe `toString` can be refactored to incorporate that logic but only for the actual nodes that are chosen?

I had a vague memory that was not an issue. Just checked the code:
```ml|
    | CustomLayout children ->
      (* When using CustomLayout, we don't want to propagate forced breaks
       * from the children up. By definition it picks the first layout that fits
       * otherwise it takes the last of the list.
       * However we do want to propagate forced breaks in the sublayouts. They
       * might need to be broken. We just don't propagate them any higher here *)
      let _ = walk (Concat children) in
      false
```
can't see any slowdown there.

cristianoc:
But I'm not sure I understood your question.
The main point of the suggestion is: replace exponential traversal with linear traversal.

cristianoc:
Take it with a grain of salt, done a quick back-of-the-envelope calculation. But the difference, given exactly the same grammar we have today, is from `3 ^ n` to `3 * n`.

leeor:
Walking the `children` using `Content` means a `fold_left` that traverses all variations. Unless I am missing something here, again, not very familiar with this code base. From my previous tests with the lazy evaluated variations, skipping the call in `toString` make it almost instantaneous.

I now brought back the `propagateForceBreaks` function and the call to it from `toString` but without processing `CustomLayout` variations. Instead, I call it from the process helper in `toString` after a `CustomeLayout` variation is chosen. This greatly impacts the overall performance, with only a minimal impact on formatting.

Performance:
```
./_build/install/default/bin/rescript -print res   0.65s user 0.02s system 46% cpu 1.452 total
```
Compared to ~22 seconds before lazy evaluation and the `propagateForceBreaks` changes.

The formatting difference (hence failed tests):
```diff
diff --git a/tests/printer/other/expected/nesting.res.txt b/tests/printer/other/expected/nesting.res.txt
index 0cafca836cfc..b25b2d1724cd 100644
--- a/tests/printer/other/expected/nesting.res.txt
+++ b/tests/printer/other/expected/nesting.res.txt
@@ -1,7 +1,5 @@
-let unitsCommands = state.units->Js.Array2.mapi((
-  {unit: targetUnit, coordinates: targetCoordinates},
-  i,
-) => {
+let unitsCommands =
+  state.units->Js.Array2.mapi(({unit: targetUnit, coordinates: targetCoordinates}, i) => {
     // n^2
     let res = []
     state.units->Js.Array2.forEachi((
@@ -50,4 +48,4 @@ let unitsCommands = state.units->Js.Array2.mapi((
       }
     })
     res
-})
+  })
```

cristianoc:
Tried the latest version here.

I've made the example slightly bigger:
```
% dune exec -- time rescript tst.txt
let foo = () =>                 
  bar(x =>
    bar(x =>
      bar(x =>
        bar(x =>
          bar(x =>
            bar(x =>
              bar(x =>
                bar(x =>
                  bar(x =>
                    bar(x =>
                      bar(x =>
                        bar(x =>
                          bar(x => bar(x => bar(x => bar(x => bar(x => bar(x => bar(x => x)))))))
                        )
                      )
                    )
                  )
                )
              )
            )
          )
        )
      )
    )
  )
        4.78 real         3.88 user         0.58 sys

```

cristianoc:
With just another 3 more nested callbacks, it uses up nearly all the memory on my machine and takes 57 seconds.
Can't optimise against exponentials, one needs to remove them.

cristianoc:
The first issue is that it generates an exponential number of docs, even before trying to format them.

cristianoc:
Turns out parsing also takes time on a bigger nested example. So that time should be factored out of the measurement.

cristianoc:
The 2 cases produced in `printArgumentsWithCallbackInLastPosition` alone creates an exponential number of docs.

cristianoc:
To summarise my understanding.
There's doc construction and doc printing.
This separation makes it difficult to express interesting logic when several layout cases need to be considered out of which one is chosen.
The only construct supporting logic is Customlayout, which only supports the logic "try this, else try that, else that".

The current PR puts the real logic behind lazy, so that printing can do more work. It has gotten rid of a 3^n explosion but there's still a 2^n explosion left at doc construction time.

Perhaps it's possible to patch this PR to remove the 2^n.
The alternative is to put more logic into the doc so that linearly size, non-lazy, docs can be contructed.

cristianoc:
One step forward, in the lazy approach, here: https://github.com/rescript-lang/syntax/pull/590
The output is still not the desired one but it's fast.

Another approach consists of not generating  at all an exponential number of docs (whether hidden behind lazy or not), but only generate those corresponding to the sequences: https://github.com/rescript-lang/syntax/pull/576#issuecomment-1164439620

leeor:
My thoughts:
1. We're already seeing a formatting change, is that acceptable?
2. I don't mind laziness, it is an easy solution to reduce upfront costs of any values we won't really need later on, so can provide value whatever the final process looks like.
3. The current state of this PR, given we're ok with (1) above, is a significant improvement in runtime. Maybe that's good enough to merge and release so that some benefit will already be made available? (also see the next point)
4. I admit I have not fully understood the change you're looking to do in `fits` and `willBreak`. Adding a `level` argument to `willBreak` and `process` all the way down to `calculate` in `fits` and selecting the second `CustomLayout` case whenever it's higher than 1, did not make any change to the runtime.

WDYT?

leeor:
By output is still not the desired one, do you mean the final formatting being printed, or timing/code structure?

cristianoc:
It's fast but it does not print what we want (e.g. the example motivating this, but also changes to the tests, need to check if they are good or not).
So if it can be made to print what we need , then we can keep it.

More experimentation needed.

cristianoc:
This prints what we need:

```ml
    | CustomLayout children ->
      let _ = List.fold_left
        (fun forceBreak child ->
          let childForcesBreak = walk (Lazy.force child) in
          forceBreak || childForcesBreak)
        false children in
      false
```

but it is slow, as it undoes all what lazy does.

Btw this is the output:
```rescript
let foo = () => bar(x => bar(x => bar(x => bar(x => bar(x => bar(x => bar(x => bar(x => bar(x =>
                    bar(x => bar(x => bar(x => bar(x => bar(x => x)))))
                  )))))))))
```

It corresponds to what TS does.

cristianoc:
> 4. I admit I have not fully understood the change you're looking to do in `fits` and `willBreak`. Adding a `level` argument to `willBreak` and `process` all the way down to `calculate` in `fits` and selecting the second `CustomLayout` case whenever it's higher than 1, did not make any change to the runtime.

The idea is to pass an extra parameter to all the functions that *produce* a doc. So all the printing functions.
And: don't print a call-back that breaks, inside another callback printed as non-breaking.
So you literally generate a smaller doc, with fewer cases.

PR #577, opened 2022-06-23
Run ocamlformat check in tests

cknitt:
As suggested by @cristianoc to increase the chance of discovering formatting issues during local development/testing (and not later on in CI).


cristianoc:
Does that mean files get overwritten when doing `make test`?

cristianoc:
Is it possible to do a little more:
If it fails explain that it was not formatted, and they can run `make format` to format it.

On `make format`, at the end say that it was reformatted. Which is more understandable than a last line saying "Promoting...`.

PR #581, opened 2022-06-24
Bench on Linux

nkrkv:
Now `make bench` works on Linux. Let it be?

Please, macOS users, check I didn’t break the macOS bench build accidentally.


cristianoc (approved):
There was a `;` missing. Works now.

cristianoc:
Thanks for providing this!

Issue #583, opened 2022-06-27
Annotations on field names are eaten by formatting.


cristianoc:
```rescript
 let make = ({name: @optional name}: props<_>) => name->Belt.Option.getWithDefault("")->React.string // 2
```

becomes

```rescript
  let make = ({name}: props<_>) => name->Belt.Option.getWithDefault("")->React.string // 2
```

Also, the punned version gives a syntax error:
```rescript
  let make = ({@optional name}: props<_>) => name->Belt.Option.getWithDefault("")->React.string // 2
```

This is also eaten by formatting:
```rescript
let z = name => { name : @optional  name, x:3}
```

And this is a syntax error:
```rescript
let z = name => { @optional name, x: 3}
```

PR #584, opened 2022-06-27
Dropped annotations from punned record labels.

cristianoc:
Add examples of expression and patterns showing where they are dropped.
Support all the cases discussed in https://github.com/rescript-lang/syntax/issues/583

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -2893,7 +2893,7 @@ and parseBracedOrRecordExpr p =
     let braces = makeBracesAttr loc in
     {expr with pexp_attributes = braces :: expr.pexp_attributes}
 
-and parseRecordRowWithStringKey p =
+and parseRecordExprRowWithStringKey p =
   match p.Parser.token with
   | String s -> (
     let loc = mkLoc p.startPos p.endPos in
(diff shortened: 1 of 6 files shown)


src/res_core.ml
```diff
@@ -2674,6 +2676,7 @@ and parseJsxChildren p =
 and parseBracedOrRecordExpr p =
   let startPos = p.Parser.startPos in
   Parser.expect Lbrace p;
+  let attrsForFields = parseAttributesForFields p in
```
cristianoc, line 2679: This is *very* close to surface syntax.

src/res_core.ml
```diff
@@ -6250,9 +6263,29 @@ and parseAttribute p =
           ] )
   | _ -> None
 
+and parseAttributeForFields p =
+  let isAttributeForField p =
+    Parser.lookahead p (function state ->
```
IwanKaramazow, line 6268: I guess the grammar forces us into using a lookahead?

cristianoc: It's super overloaded over there. Apparently `{JSX stufff}` only works because the `JSX` attribute drives the parser into a specific path (which does not consider records). Which itself is arbitrary.

I'm thinking of considering proper surface syntax early on, so one does not need to distinguish based on the name of attributes, which is really not robust.

cristianoc: Before this PR, it already distinguished based on the *presence* of attributes, which is also unclear whether intended.

PR #586, opened 2022-06-27
JSX PPX V4 part2

mattdamon108:
* [ ] the new JSX transform with ~~`Js.React`~~ `Jsx` in ~~rescript-react~~ the compiler
    * [ ] https://github.com/rescript-lang/rescript-compiler/pull/5484
    * [ ] https://github.com/rescript-lang/rescript-react/pull/49
* [x] Wired up to the new configuration in bsconfig.json
    * [ ] https://github.com/rescript-lang/rescript-compiler/pull/5484
* [ ] Polish the loc properly
    * [x] props loc
    * [x] Upper case application doesn't point to the definition
* [x] Test for newtype
* [x] Test for typeConstraints
* [x] `forwardRef` interface
* [ ] Engage the feedback from the forum
* [x] file-level config `@@jsxConfig( ... )`


cristianoc:
After this https://github.com/rescript-lang/syntax/pull/588, assuming all goes well, the change required will be simply to use `@ns.optional` instead of `@optional` which will be removed.

cristianoc:
And this converts to `@ns.optional` including the JSX ppx: https://github.com/rescript-lang/syntax/pull/589.
Would you take a look and see if you find any issues.

cristianoc:
Would you rebase on master now that the surface syntax has landed.

cristianoc:
Is the idea that clients of the ppx, such as rescript-react, will define their types using`Js.React....`?
I guess we need to write a little spec for what those types are, so client libraries can use them.

mattdamon108:
> Is the idea that clients of the ppx, such as rescript-react, will define their types using`Js.React....`?

Frankly, I didn't think that far 😄, but it can be used by clients. `Js.React` is just for adding the new jsx transform bindings without breaking or updating the rescript-react at this stage. In order for type checker to check types without failure, `Js.React` needs to vendor the basic types such as `Js.React.element`, `Js.React.domProps`, etc. Especially, `Js.React.domProps` will be used to refactor v3 against `type ReactDOMRe.domProps`.

> I guess we need to write a little spec for what those types are, so client libraries can use them.

Sure, I'll write down on it.

cristianoc:
I thought this new jsx mode was opt-in, so off by default even on users of V4.
If true, this might in fact be the best situation to introduce a "breaking change" in that a newer version of rescript-react needs to be used.
Is this correct?

cristianoc:
I think the new mode should have no reference to rescript-react at all. But the references should be the other way round: rescript-react (and others) can use the types defined in this PR (and the corresponding one in the compiler that adds types under Js.React).

Btw, instead of `Js.React`, perhaps `Js.Jsx` is better namespace.

cristianoc:
Made some comments on the compiler PR, but realised it might be simpler to discuss directly on the spec once it is written.
So feel free to ignore those comments for now.

mattdamon108:
I've updated the spec in https://github.com/rescript-lang/syntax/pull/586/commits/75f8a284d8d1ab71a7efd010dd9a6f27baa73170.

> I thought this new jsx mode was opt-in, so off by default even on users of V4.
> If true, this might in fact be the best situation to introduce a "breaking change" in that a newer version of rescript-react needs to be used.
> Is this correct?

I agreed that the new jsx mode is better off by default even for the users of V4. I'm not clear about the direction of the core team regarding the rescript-react. I think it is the only choice to add `Js.React` and the bindings of the new jsx mode into the compiler if we want to keep the rescript-react from updating and breaking. Furthermore, it would be better design choice to opt in the jsx bindings and types. It is not a big deal in consideration that the jsx ppx is already sitting inside the compiler.

But, if the core team is wiling to update the rescipt-react and keep it as one of dependencies, then I can make PR for rescript-react for the new jsx mode.

cristianoc:
> I've updated the spec in [75f8a28](https://github.com/rescript-lang/syntax/commit/75f8a284d8d1ab71a7efd010dd9a6f27baa73170).
> 
> > I thought this new jsx mode was opt-in, so off by default even on users of V4.
> > If true, this might in fact be the best situation to introduce a "breaking change" in that a newer version of rescript-react needs to be used.
> > Is this correct?
> 
> I agreed that the new jsx mode is better off by default even for the users of V4. I'm not clear about the direction of the core team regarding the rescript-react. I think it is the only choice to add `Js.React` and the bindings of the new jsx mode into the compiler if we want to keep the rescript-react from updating and breaking. Furthermore, it would be better design choice to opt in the jsx bindings and types. It is not a big deal in consideration that the jsx ppx is already sitting inside the compiler.
> 
> But, if the core team is wiling to update the rescipt-react and keep it as one of dependencies, then I can make PR for rescript-react for the new jsx mode. I think that releasing the V4 is the best chance to update the rescript-react.

Definitely rescript-react can, and should be updated.
The only comment is: it must be possible for people to have a smooth upgrade path without updating everything. As they will have dependencies that they cannot upgrade as they're not the maintainers.
And they will have product code that they need to migrate gradually.
So I think a smooth upgrade path is one where the only thing you update is the V4 ppx. Everything still works. You convert little by little.
Once all is done, flip the switch and upgrade rescript-react.

Does it make sense?
I haven't specified what "flip the switch" means. It's open for design. One possibility would be: turn on the new jsx.

cristianoc:
I think we need to explore the gradual upgrade path right now. As that affects the design of the various ppx options.
How does that sound?

cristianoc:
About being React-specific: here's a TS page:
https://www.typescriptlang.org/docs/handbook/jsx.html

Notice in particular: "[JSX](https://facebook.github.io/jsx/) is an embeddable XML-like syntax. It is meant to be transformed into valid JavaScript, though the semantics of that transformation are implementation-specific. JSX rose to popularity with the [React](https://reactjs.org/) framework, but has since seen other implementations as well"

I think it makes sense to aim for the same level of generality.

Notice it says "TypeScript ships with three JSX modes" and not "TypeScript ships with three React modes".

Hence my suggestion to use the `Js.Jsx` namespace not `Js.React`.

nkrkv:
> Btw, instead of `Js.React`, perhaps `Js.Jsx` is better namespace.

This! A language should outlast a framework. The community puts some effort trying to use ReScript with SolidJS, Remix, etc. And it will succeed one day I think. Hardcoding the “React” name into the language introduces an unnecessary and incorrect illusion that ReScript is *only* for React front-end apps.

mattdamon108:
> I think we need to explore the gradual upgrade path right now. As that affects the design of the various ppx options. How does that sound?

Totally, it makes sense. I got your point. It is a good idea to explore the gradual upgrade path first, then fix the spec as the paths. Let me write some of the possible paths with the following specs.

cristianoc:
Perhaps one way to go is to do 2 of these,  the easiest being React and React Native.
For React Native, one currently needs to install both rescript react, and rescript react native. A goal would be to be able to only install rescript react native, i.e. the part that adds bindings from actual components. I think for doing that, one needs to move the relevant types inside the compiler.

So one could, for example, have the core JSX types in `Js.Jsx`. Then both rescript react, and rescript react native, as independent packages, can make use of those types and add whatever is specific to the libraries.

This is just a though. By working out the details one can probably get a good idea.

I think one next step is to extent the spec with the types: what types do `jsx` and `jsxDom` have? One might be able to give them types that are independent from the specific framework. Or at least try, see if it makes sense.

cli/JSXV4UPGRADE.md
```diff
@@ -0,0 +1,44 @@
...
+   ```json
+   "react": {
+     "jsx": 3,
+     "runtime": "classic"
+   }
+   ```
+
+   The `react.runtime` affects nothing, no matter of `"classic"` or `"automatic"`.
+
+   b. JSX V4 with classic mode
```
cristianoc, line 20: This doc looks good.
One aspect to consider more is how to adopt `"jsx": 4` gradually. This `bsconfig.json` change is sufficient when the entire project is using V4. But we need to also think about what to do when half the project uses V3 and half the project uses V4.

cristianoc: At the very least one needs to specify per-file `"jsx": 4` vs `"jsx": 3` i.e. override the bsconfig locally.
In fact one might need to be able to do that in the middle of a file.
E.g. What if the file is `<A > <B /> </A>` and:
1 initially both `A` and `B` are in V3, and the current file is using V3
2 make the current file use V4
3 make `A.res` use V4
4 make `B.res` use V4

cristianoc: So at point 2, the jsx would cal `<A>`  the V4 way but it expects to be called the V3 way.
I think this is solved by turning on V4 at the top of the file. And locally turning on V3 where `<A > <B /> </A>` is.

cristianoc: For 3, `A` uses V4 but `B` is still on V3. 
We need to be able to turn on V3 inside `<A > ... </A>` so that we can call `B` with V3.

cristianoc: For 4, we don't need anything anymore. Just turn on V4 at the top of the current file.
When every file has been done like that and the project is converted, then non per-file config is needed anymore.

Does this make sense?
Is it going to work?

cristianoc: If this is what we need, then the local config needs to go into an attribute. As attributes can both be used module-level (for the top of the file) or at the level of individual expressions.

cristianoc: This is how it might look like. Notice the use of module-level attributes `@@` and expression-level `@`.

### 1. ###
```rescript
<A> <B/> </A>
```

### 2. ###
```rescript
@@jsxconfig(4)

...

@@jsxconfig(3)

<A> <B/> </A>
```

### 3. ###
```rescript
@@jsxconfig(4)

...

<A> @jsxconfig(3) <B/> </A>
```

### 4. ###
```rescript
@@jsxconfig(4)

...

<A> <B/> </A>
```

mattdamon108:
> I think one next step is to extent the spec with the types: what types do `jsx` and `jsxDom` have? One might be able to give them types that are independent from the specific framework. Or at least try, see if it makes sense.

I think I got your point. I thought `Js.React` is only for V4 + new JSX mode workaround actually. I think I'm aligned to your intention now. I'll give it a try in that way.

cli/JSXV4UPGRADE.md
mattdamon108: This is a really good point I didn't consider enough frankly. Can you elaborate on the case that A, B are using V3 against the current file? I can imagine that A, B are all installed dependencies which has the bsconfig.json respectively. If I'm correct, I'm afraid that it is not available to use the new JSX mode in`Js.Jsx` for the first release of V4. Because A, B have types from rescript-react, e.g. `React.element`, but the current file will have a type based on `Js.Jsx.element`. Those are not mixed to compose the components as flexible as users want. I mean that the new JSX mode is not available without breaking the existing projects by combining the V3 components. IMO, the first release of V4 is available 1 ~ 2.c in my upgrade doc. Definitely we have to focus and test thoroughly the 2.b is not going to break any projects with V3 dependencies. I think we can achieve that with current implementation. Anyway, 2.c is just the experimental feature for the projects without V3 dependencies.

mattdamon108: And for the next stage, we can make the `Js.Jsx` matured as an experimental feature not only for the new JSX mode, so that the external dependencies are going to adopt and upgrade their modules, especially rescript-react. I think we can ship the V4 with "automatic" runtime by default later. IMHO, I'm afraid that the tweaks such as `@jsxconfig` may cause complexity issues and confusion for users.

cristianoc: I think the process is:
- switch to V4 but do *not* use the new jsx mode. So everything is backwards compatible
- gradually convert the components to V4
- eventually use the new JSX mode if you like
Is that what you have in mind?

It's important that `A` and `B` be just files (or even local modules) not entire dependencies, so you can gradually switch over a large application. Sure most of the times you can just switch them over all at once, but not always.

cristianoc: The majority of cases will be, just switch to V4 and there's nothing to do. But not in all cases. I think.

cristianoc: I think the relevant types are things like:
```rescript
type element

type componentLike<'props, 'return> = 'props => 'return

type component<'props> = componentLike<'props, element>

/* this function exists to prepare for making `component` abstract */
external component: componentLike<'props, element> => component<'props> = "%identity"

```

In there the only "secret" type is `element`.
So in order for things to work all that is needed is that `React.element` is recognised by the type system as the same type as `Js.Jsx.element`. There are many ways one can achieve that. It's easy to define modules and open them locally to shadow type definitions.

One can define `Jsx.element` to be `React.element`. Or use `Js.React.element` defined in that way.
Or one can `open` use a module called `React` that defines `type element = Js.Jsx.element` etc etc.

cristianoc: So I'd like firs to know what the primitive types are. What the primitive values are (`jsx` and `jsxDom` etc)) with their types.
Then *separately* see how to connect that to rescript-react.
And to rescript react native.
And ideally to something else too.

cristianoc: So back to the `A` and `B` components, V3 vs V4 scenario.
It really depends on the nature of projects how detailed one needs to be.
The happy case is: turn on V4 and do nothing. It works.
But worries are: what if one has written some `make` functions by hand? They need to change. What if someone is exporting components to JS and then calling make directly. That needs to change too. Etc.
Since these worries are highly project-dependent, there's no way to know in advance.

Perhaps a good way to go is to document what is the worst case scenario (need to switch between V3 and V4 at arbitrary points), but don't implement it yet. Just do some tests.
E.g. take a few large apps and see what breaks, if anything.
Based on that, decide how deeply to go.

mattdamon108: > E.g. take a few large apps and see what breaks, if anything.

There are two apps for the good battle-test, my company apps 😏. It is quite large, I think.

cristianoc: Great plan. Looking forward to hearing how it goes.

cristianoc: Kind of depends on what we need. Just knowing most things are possible frees up the design space.
We don't need to go into detail with one until we know it's needed.

Proposal: how about you just write it for React. So use `React.element` etc. Then I'll have a go at the general version.

mattdamon108: > Proposal: how about you just write it for React. So use `React.element` etc. Then I'll have a go at the general version.

OK, I'll give it a try!

cli/JSXV4UPGRADE.md
```diff
@@ -0,0 +1,44 @@
...
+   a. JSX V3
+
+   ```json
+   "react": {
+     "jsx": 3,
+     "runtime": "classic"
+   }
+   ```
+
+   The `react.runtime` affects nothing, no matter of `"classic"` or `"automatic"`.
```
ryyppy, line 18: If the `react.runtime` flag "affect nothing", why is it there?

cli/JSXV4UPGRADE.md
```diff
@@ -0,0 +1,44 @@
...
+   c. JSX V4 with new JSX mode (Experimental)
+
+   ```json
+   "react": {
+     "jsx": 4,
+     "runtime": "automatic"
+   }
+   ```
+
+   JSX V4 with `Js.React` which needs the peer-dependecy of React v17.\* or higher. It may break the project with dependencies which are using the explicit types of `rescript-react`.
```
ryyppy, line 38: > `Js.React`

What's the idea of making it a submodule of `Js`? How is a user supposed to define this binding?
Also, do we again couple the JSX implementation to React? I thought we wanted to explore more liberate systems that will allow different targets.

mattdamon108: It is just a test try to add missing bindings of the new JSX mode which is wrongly bound. https://github.com/rescript-lang/syntax/pull/547#issuecomment-1166177098

cli/JSXV4UPGRADE.md
```diff
@@ -0,0 +1,44 @@
+**JSX V4 upgrade path**
+
+1. No upgrading rescript-react && no changes in bsconfig.json
+
+   JSX V3 by default.
```
ryyppy, line 5: Unless we did any changes in the compiler, if there's no `react` setting in your `bsconfig`, no JSX transformation will be applied.

So either that sentence is wrong, or I missed the change in the compiler setting.

cli/JSXV4UPGRADE.md
mattdamon108: Good catch. It doesn't have to be there, but as an aspect of schema, it works with `react.jsx: 4`

cli/JSXV4UPGRADE.md
ryyppy: I guess what I rather wanted to see is a config model that is similar to [TS jsx option](https://www.typescriptlang.org/tsconfig#jsx)

cli/JSXV4UPGRADE.md
mattdamon108: Ah, I made a PR in the compiler for the new settings.

cli/JSXV4UPGRADE.md
mattdamon108: Thanks!

cli/JSXV4UPGRADE.md
mattdamon108: I tried to test changing the type declaration in `React.res` of rescript-react in the node_modules.
```rescript
type element = Js.React.element
```
It seems working with composing V4 + new JSX mode & V3 in the same project. But I can't figure out how to open and shadow the type declaration after making the `React` module in my project. Technically I can make the `React.res` in my project, but It seems not possible to open `React` inside it. Is it what you meant?

image

cristianoc: Actually I quite like Patrick's proposal of being super lightweight.
So the ppx just emits calls to `Jsx.jsx` and `Jsx.jsxDom`, and the compiler defines nothing at all.
Then RescriptReact needs to provide a module called `Jsx` instead of `React`, and that's all.

cristianoc: So rescript-react can use whatever type it wants for `jsx` and `jsxDom` as long as the code generated by the ppx type checks.

cristianoc: > If so, rescript-react needs to be upgraded. Because it’s binding for jsx is not right. That is all the way back to thr begining of our duscussion.

Sorry for the confusion. I guess there are too many things going at the same time and it's very easy to misunderstand.
I was under the impression that it's perfectly OK to rely on a new version of rescript-react to support the new jsx mode.
But it's essential that people can switch to V4 (not with the new jsx) without changing anything.
I thought that was the plan.

I think once people have migrated to V4, then it's not a problem to upgrade rescript-react and turn on the new jsx mode.

cristianoc: Just one minor comment. It's not "they". It is "us".
As long as people are kept in the loop and questions asked in the forum when in doubt, we can just go ahead make new version of rescript-react whenever needed.

mattdamon108: It's you or Patrick's call, I can make PR for new `Jsx` module if you guys want. Or I'll await the `Jsx` module's interface are going to be fixed, once it is fixed, PPX V4 just emits to call that. I guess V4 is almost done now.

cristianoc: Definitely a PR for the new Jsx is useful for testing. Making sure that all works.
Then make a release when the time is right.

cristianoc:
I have update the upgrade docs based on the discussions. Let me know if it makes sense.

cli/JSXV4UPGRADE.md
```diff
@@ -1,44 +1,40 @@
...
-   "react": {
-     "jsx": 4,
+   "jsx": {
+     "version": 4,
      "runtime": "automatic"
    }
    ```
 
-   JSX V4 with `Js.React` which needs the peer-dependecy of React v17.\* or higher. It may break the project with dependencies which are using the explicit types of `rescript-react`.
+   JSX V4 with `"automatic"` runtme needs React v17.\* or higher as a peer dependency. It will require a new version of`rescript-react`.
```
mattdamon108, line 38: Totally make sense!

mattdamon108:
I've added the proposal `Jsx` module in the rescript-react in my local project, and fixed the ppx to emit a call to it. It seems perfectly working. I'm fixing a small thing about the discouraged forwardRef example now. Once it is done, I'll make my proposal `Jsx` PR to rescript-react.

cli/reactjs_jsx_ppx_v4.ml
```diff
@@ -1064,7 +1064,8 @@ let transformComponentDefinition nestedModules mapper structure returnStructures
           @
           if hasForwardRef then
             [
-              (Location.mknoloc (Lident "ref"), Pat.var (Location.mknoloc "ref"));
+              ( Location.mknoloc (Lident "ref"),
+                Pat.var ~attrs:optionalAttr (Location.mknoloc "ref") );
```
cristianoc, line 1068: Before continuing adding `mknoloc` that later need to be changed anyway.
The loc should be corresponding to something in the original source. The only question is whether the `loc_ghost` field is set to true or false.
(It would be set to true exactly once for each location used in the code produced by the ppx).

mattdamon108: Good catch. I'm gonna polish the loc corresponding to original source code.

mattdamon108:
Updated the V4 upgrade doc as per https://github.com/rescript-lang/rescript-compiler/commit/81029b30f20e8b12b9dff1229228e31daa5e7c5b

mattdamon108:
fix as per the review https://github.com/rescript-lang/syntax/pull/586/commits/b55f739611a85ced86adb6ecc5246b5bed440e00 https://github.com/rescript-lang/syntax/pull/586/commits/0e83018eca7f0e8b59074e3e635fc40ddd3f85c9

cristianoc:
@mattdamon108 what is the scope of the file-level jsx config update?
What if I update the config in a submodule, what happens after that submodule? Does the update stay or not?

cristianoc:
The simple possible semantics is it stays for the rest of the file, and there's nothing special about submodules.
If that's the intention, the only thing to do is to check that these annotations are visited in the expected order, i.e. the line order. And I would guess this is in fact the order in which they are visited.

mattdamon108:
I'm not sure I understand your context correctly.
Yes, the update of the config with this top-level attribute stays for the rest of the file. It doesn't affect the submodules. I'm not sure you already acknowledge, that the top-level attribute is only working in case it is written in the first of the line of a file. Am I standing in the same context?

cristianoc:
> not sure you already acknowledge, that the top-level attribute is only working in case it is written in the first of the line of a file
That is not true.

cristianoc:
Where is the implementation? If it works only at top-level I'll suggest a quick fix to make it work everywhere.

mattdamon108:
Currently, it is working as below.
```rescript
// A.res
@@jsxConfig({version: 4}) // first line of A.res
...
```
The whole structure_items in A.res is transformed with V4.

```rescript
// A.res
...
@@jsxConfig({version:4 }) // somewhere in the middle of A.res
...
```
The whole structure_items in A.res is transformed with the config by bsconfig.json.

Can you explain what `it works everywhere` means?

cristianoc:
Everywhere means: including nested modules.

mattdamon108:
Ah! I got what you meant now. Yes, it is working accordingly!

cristianoc:
> Ah! I got what you meant now. Yes, it is working accordingly!

Perfect. I will clarify this in the spec.

cristianoc:
In fact, this is just communication overhead.
Let me take a look at the implementation and spec, will make sure that agree.

mattdamon108:
Okay.
The compiler `ppx_entry.ml` will call `reactjs_jsx_ppx.ml` with config in bsconfig.json. `reactjs_jsx_ppx.ml` try to find `@@jsxConfig` in the code(structure_item list). If the first structure_item is `@@jsxConfig`, then `reactjs_jsx_ppx.ml` update the config against bsconfig.json. After updating, it will call `reactjs_jsx_ppx_v3.rewrite` or `reactjs_jsx_ppx_v4.rewrite` respectively for the all of structure_items.

cristianoc:
I think the ppx better be moved into the compiler.
So it's directly accessible by bsc. No command-line options.

mattdamon108:
I would suggest `@@jsxConfig` has its scope for the file only. What's your intention? for the whole project? updating the project's jsx config with the single line of `@@jsxConfig`? IMHO, is it too much?

mattdamon108:
If so, I'm afraid that the order of file would be a problem. There would be discrepancies between files before and after the compiler encounters the `@@jsxConfig`.

cristianoc:
> I would suggest `@@jsxConfig` has its scope for the file only. What's your intention? for the whole project? updating the project's jsx config with the single line of `@@jsxConfig`? IMHO, is it too much?

No that's not sufficient. Let me just fix this. I'll update the PR.

cristianoc:
> If so, I'm afraid that the order of file would be a problem. There would be discrepancies between files before and after the compiler encounters the `@@jsxConfig`.

No it's completely declarative.

mattdamon108:
The implementation is here https://github.com/mattdamon108/syntax/blob/14a8d8de0460e53395ab9b87b598e6043ae0afd4/cli/reactjs_jsx_ppx.ml#L80

cristianoc:
Did some of it here: https://github.com/rescript-lang/syntax/pull/603
Need to go now but can add more later.
Looks like V3 and V4 need to be fused, so either one can be called depending on what the mode is.

cristianoc:
It's not finished yet. It runs either one mapper or the other one.

cristianoc:
Now it should be substantially done, modulo some bugs.

cristianoc:
I think the granularity is not enough yet, as it cannot switch between V3 and V4 inside one structure.
Need to go down to the structure item (or signature item) level.

mattdamon108:
> Also need a test where the config changes in inner modules.

I have a quick question about the scope and level of the config update.
```rescript
@@jsxConfig({version: 4})

let make = () => body // 1

module Foo = {
  @@jsxConfig({version: 3})
  module FooA = {
    let make = () => body // 2
  }
}

let make = () => body // 3
```
Regarding the semantics, definition 3 needs to be transformed with V4, but It is transformed with V3 now. Do we need some kind of level such as one in the typecore in the compiler?

cristianoc:
Done.

cristianoc:
Other than that, the ppx should be reviewed and check that it does not allocate more than it needs to.

mattdamon108:
> Other than that, the ppx should be reviewed and check that it does not allocate more than it needs to.

Do you mean the partial config update due to mis-location?

cristianoc:
> > Other than that, the ppx should be reviewed and check that it does not allocate more than it needs to.
> 
> Do you mean the partial config update due to mis-location?

The jsx transform happens on every file compiled, so inefficiencies there multiply easily.

cristianoc:
Maybe the efficiency is OK as it is, but it does not hurt doing one pass of thinking about what it does.
One simple thing would be to check if a transformation needs to be applied, and only do it in that case.
E.g. if there are no uses of JSX or component definitions then one can return the original AST directly instead of traversing and copying everything.

cristianoc:
(I did one final small cleanup -- finished now. Will not touch the code again back to you)

mattdamon108:
> The jsx transform happens on every file compiled, so inefficiencies there multiply easily.

> Maybe the efficiency is OK as it is, but it does not hurt doing one pass of thinking about what it does.
> One simple thing would be to check if a transformation needs to be applied, and only do it in that case.
> E.g. if there are no uses of JSX or component definitions then one can return the original AST directly instead of traversing and copying everything.

Isn't it inevitable? That is how the ppx works, I think. Maybe we can add more semantics, such as something like `@@noNeedJsxTransform`. But it is too much for users, probably efficiency is not the users' concern in these case.

cristianoc:
> > The jsx transform happens on every file compiled, so inefficiencies there multiply easily.
> 
> > Maybe the efficiency is OK as it is, but it does not hurt doing one pass of thinking about what it does.
> > One simple thing would be to check if a transformation needs to be applied, and only do it in that case.
> > E.g. if there are no uses of JSX or component definitions then one can return the original AST directly instead of traversing and copying everything.
> 
> Isn't it inevitable? That is how the ppx works, I think. Maybe we can add more semantics, such as something like `@@noNeedJsxTransform`. But it is too much for users, probably efficiency is not the users' concern in these case.

It's not inevitable. Well using a traditional ppx it is but this one is inside the compiler so you get full control.
One can literally do a quick zero-allocation pass to see if a transform is needed. And if it is not, do nothing.
Such a pass would use an ast iterator instead of an ast mapper, and keep a boolean reference to see if something was found that will require transformation.

cristianoc:
I'm not suggesting to do that now. Just, to have a quick look at what the ppx does and see if there are some obvious inefficiencies.

mattdamon108:
I'm going to plug it into the compiler and move forward. 👍

cli/reactjs_jsx_ppx.ml
```diff
@@ -2054,7 +2061,7 @@ module V4 = struct
         let retPropsType =
           Typ.constr ~loc:pstr_loc
             (Location.mkloc (Lident "props") pstr_loc)
-            (makePropsTypeParams namedTypeList)
+            (makePropsTypeExternal namedTypeList)
```
cristianoc, line 2064: Why does an external definition need special treatment?

mattdamon108: Wow, you're a quick reviewer. Actually, it should be the same as the `makePropsTypeExternal`, so I'm just about to merge and rename those functions.

cristianoc:
This is getting to a more stable state now.
For backwards compatibility with V3 one would like to see no changes in test results. And this PR currently mixes a number of changes to existing tests.
How about leaving existing tests alone, and just add new ones for the new functionality.

Or even split up this PR at some point.

cristianoc:
Also, I would like to have a go at cleaning up location info, but for that I would wait until the code is pretty much done.
Are there any major changes coming?
Also, are V3 and V4 going back to separate files or stay as they are right now?

cli/JSXV4UPGRADE.md
```diff
@@ -0,0 +1,43 @@
...
+
+   ```json
+   "jsx": {
+     "version": 3,
+     "module": "react",
+     "mode": "classic"
+   }
+   ```
+
+   **Note:** When using `jsx` v3, the `runtime` option will be ignored.
```
cristianoc, line 19: this line seems old

cli/JSXV4.md
```diff
@@ -98,3 +126,15 @@ function has the name of the enclosing module/file.
...
 ReactDOMRe.createElement(ReasonReact.fragment, [comp1, comp2, comp3])
 ```
+
+**File-level config**
+
+The top-level attribute `@@jsxConfig` is used to update the jsx config for the rest of the file (or until the next config update). Only the values mentioned are updated, the others are left unchanged.
+
+```rescript
+@@jsxConfig({version: 4, mode: "automatic"})
+// The jsx config is updated for the rest of the file.
```
cristianoc, line 136: this also seems old -- needs a bit more explanation of what is possible

cli/JSXV4UPGRADE.md
```diff
@@ -0,0 +1,43 @@
...
+
+2. New opt-in configuration added in bsconfig.json
+
+   a. JSX V3
+
+   ```json
+   "jsx": {
+     "version": 3,
+     "module": "react",
+     "mode": "classic"
```
cristianoc, line 15: why even mention `mode` here?
and why even mention `react` as V3 will only work with react?

mattdamon108:
> Also, I would like to have a go at cleaning up location info, but for that I would wait until the code is pretty much done.
> Are there any major changes coming?

Yeap, next thingis the missing implementation for `newType`. I don't think it would take much time to implement, after that I'm gonna clean up the whole location usage from top to bottom. Recently, I haven't got enough time to work, but I try to speed up this week.

> Also, are V3 and V4 going back to separate files or stay as they are right now?

No sperate. I think that separated V3 and V4 modules in same file are good to go.

mattdamon108:
> How about leaving existing tests alone, and just add new ones for the new functionality.
> 
> Or even split up this PR at some point.

Got it. I think we don't need to make new PR for tests, because I don't think I have more things to change the existing tests for now. Like above I said, I have the newtype implementation left which is going to change the tests for V4.

cli/JSXV4UPGRADE.md
mattdamon108: Wow, it looks old now. Docs are easy to be outdated.

cli/JSXV4.md
mattdamon108: add some explanations there

cli/JSXV4UPGRADE.md
mattdamon108: Thanks. removed.

cli/JSXV4.md
```diff
@@ -132,9 +132,24 @@ ReactDOMRe.createElement(ReasonReact.fragment, [comp1, comp2, comp3])
...
+
+  @@jsxConfig({ version: 4, mode: "classic" })
+
+  module R2 = {
+    @react.component  // V4 with `React.createElement`
+    let make = () => body
+  }
+}
+
+@@jsxConfig({ versino: 3 })
```
cristianoc, line 151: typo: version

mattdamon108: Thanks! fixed.

mattdamon108:
It seems silly question at this moment, but one small question came to mind why all the attributes are removed here. V3 https://github.com/mattdamon108/syntax/blob/5ccab4f20e08616690d081cab05f2e36ed28b027/cli/reactjs_jsx_ppx.ml#L1121 and V4 https://github.com/mattdamon108/syntax/blob/5ccab4f20e08616690d081cab05f2e36ed28b027/cli/reactjs_jsx_ppx.ml#L2460
First possible attribute is `@module`, is it okay to being removed?

cristianoc:
> It seems a silly question at this moment, but one small question came to mind why all the attributes are removed here.
> 
> * V3 https://github.com/mattdamon108/syntax/blob/5ccab4f20e08616690d081cab05f2e36ed28b027/cli/reactjs_jsx_ppx.ml#L1121
> * V4 https://github.com/mattdamon108/syntax/blob/5ccab4f20e08616690d081cab05f2e36ed28b027/cli/reactjs_jsx_ppx.ml#L2460
> 
> What about `@module`, is it okay to being removed?

Mostly because nobody ever noticed. The same for locations.

The truth about PPXs in general is that they don't really compose well with one another. And one tends to only use 1 ppx at any given point.

But that throwing away attributes is going to break all sorts of innocent things, such as for example what reanalyze uses.

cristianoc:
> > It seems a silly question at this moment, but one small question came to mind why all the attributes are removed here.
> > 
> > * V3 https://github.com/mattdamon108/syntax/blob/5ccab4f20e08616690d081cab05f2e36ed28b027/cli/reactjs_jsx_ppx.ml#L1121
> > * V4 https://github.com/mattdamon108/syntax/blob/5ccab4f20e08616690d081cab05f2e36ed28b027/cli/reactjs_jsx_ppx.ml#L2460
> > 
> > What about `@module`, is it okay to being removed?
> 
> Mostly because nobody ever noticed. The same for locations.
> 
> The truth about PPXs in general is that they don't really compose well with one another. And one tends to only use 1 ppx at any given point.
> 
> But that throwing away attributes is going to break all sorts of innocent things, such as for example what reanalyze uses.

Just to double check thoug: in addition to `@react.component` there could be other attributes. The correct thing to do is to preserve those attributes whenever possible. Though make sure not to make many copies: each attribute in the original code should appear exactly once in the generated code after the ppx.

I'm not even sure that it's a good idea to remove `@react.component`. As the JSX ppx is not going to run again anyway, and other tools might be able to make good use of that information, is it's the only reliable way to know if the code was generated by the JSX ppx vs handwritten. As a concrete use case: generation of interface files. Currently there are heuristics trying to understand if the code went through the JSX ppx, then try to undo the transform, and try to generate a declaration for the `.resi` file.

mattdamon108:
Yes. IMO, I agree that attributes don't need to be removed at all. I hardly come up with a reason to remove the attributes from the original AST. I'm just wondering and better to check the history with whoever knows about it. There should be a reason. But, anyway, I'll fix it and run some tests to see if any unexpected issues might be thrown.

cristianoc:
We could aim again to get this merged and continue / fine tune in a separate PR.
What needs to happen to that, is to check that backwards compatibility is preserved.

mattdamon108:
> We could aim again to get this merged and continue / fine tune in a separate PR. What needs to happen to that, is to check that backwards compatibility is preserved.

Good! 👍

cli/reactjs_jsx_ppx.ml
```diff
@@ -2590,26 +2587,6 @@ module V4 = struct
             module name.")
     [@@raises Invalid_argument]
 
-  let signature ~config mapper items =
```
cristianoc: Interesting that this was not reported by dead code analysis.

mattdamon108: Probably, I could catch those dead codes, because I still can't run the reanalyzed in my local env. I've opam switched to a different name, not default, but still error 😢 
```
❯ make reanalyze
dune build
reanalyze.exe -set-exit-code -all-cmt _build/default -suppress testrunner,compiler-libs-406 -exclude-paths compiler-libs-406 
Fatal error: exception Cmi_format.Error(_)
make: *** [reanalyze] Error 2
```

cristianoc: It's run on CI which would fail if the dead code was found.

cristianoc: I'm pretty sure I remove the other 2 copies because dead code analysis told me to.

cristianoc: And forgot about these 2.

cristianoc: > Probably, I could catch those dead codes, because I still can't run the reanalyze in my local env. I've opam switched to a different name, not default, but still error 😢
> 
> ```
> ❯ make reanalyze
> dune build
> reanalyze.exe -set-exit-code -all-cmt _build/default -suppress testrunner,compiler-libs-406 -exclude-paths compiler-libs-406 
> Fatal error: exception Cmi_format.Error(_)
> make: *** [reanalyze] Error 2
> ```

- opam install reanalyze
- remove _build
- rebuild

This has worked for others in the same situation.

cristianoc: Assuming the opam switch is a recent compiler version.

mattdamon108: > It's run on CI which would fail if the dead code was found.

Ah~ nice.

> Assuming the opam switch is a recent compiler version.

I guess there's something else in my machine. 😢 
image

mattdamon108: Actually, I found those dead codes when I tried to build the compiler with the syntax codes after copy & paste.

cristianoc: > > It's run on CI which would fail if the dead code was found.
> 
> Ah~ nice.
> 
> > Assuming the opam switch is a recent compiler version.
> 
> I guess there's something else in my machine. 😢 image

If this is windows, probably some stuff that is put in `_build` and does not belong to the project compilation.
One would need to point reanalyze more precisely than `_build/default` for example, if that extra stuff is put inside `default`.
But no need to worry about that unless interested.

mattdamon108: I'm using the m1 macbook.Alright, I'll move on

cristianoc:
Let me know when this is ready for another round of review.
Meanwhile, master is now being used for compiler `v10.1`, so you will need to rebase on that. (Most importantly the compiler repo which has new features merged -- the syntax did not change much).

mattdamon108:
> Let me know when this is ready for another round of review. Meanwhile, master is now being used for compiler `v10.1`, so you will need to rebase on master. (Most importantly rebase is needed for the compiler repo which has new features merged; the syntax did not change much).

I made rebase to master. Actually, this PR is ready for another round of review.

mattdamon108:
Nothing special. I've focused several issues since jsx config update.
* loc
* forwardRef
* more concrete props type

I think the forwardRef is less important than others, because it is discouraged usage anyway only for backward compatibility.

mattdamon108:
I'm going to test it on my company project, and let you know. I've noticed one thing hovering information in the sample project I test on. The information from V3 looks more informative than V4 in terms of props type detail.

V3
Screen Shot 2022-07-19 at 1 33 10 AM

V4
Screen Shot 2022-07-19 at 1 32 20 AM

In my guess, it needs to be done in lsp server. Is there way to show the label declarations of props type in V4?

cristianoc:
> I'm going to test it on my company project, and let you know. I've noticed one thing hovering information in the sample project I test on. The information from V3 looks more informative than V4 in terms of props type detail.
> 
> V3 Screen Shot 2022-07-19 at 1 33 10 AM
> 
> V4 Screen Shot 2022-07-19 at 1 32 20 AM
> 
> In my guess, it needs to be done in lsp server. Is there way to show the label declarations of props type in V4?

Yes there's the question of how best to represent the type information. Definitely the type definition would look better if it was expanded? Not sure it matters, but does the type definition have dummy location or actual location info?

zth:
Side note, and just vaguely related (read: please carry on, don't mind me) - it would be nice to "prettify" component hovers in general. TS also shows that in a very crude format, and it has always bothered me in both TS and ReScript. Maybe we could special case JSX component hovers somehow.

The exact format is ofc up for debate, but I'd much rather see something like this on a hover of a JSX component:

```
<Foo 
  name: string
  otherProp: int=?
/>
```

cristianoc:
> Side note, and just vaguely related (read: please carry on, don't mind me) - it would be nice to "prettify" component hovers in general. TS also shows that in a very crude format, and it has always bothered me in both TS and ReScript. Maybe we could special case JSX component hovers somehow.
> 
> The exact format is ofc up for debate, but I'd much rather see something like this on a hover of a JSX component:
> 
> ```
> <Foo 
>   name: string
>   otherProp: int=?
> />
> ```

That would happen for free if that was the way one writes the actual type definition.

cristianoc:
To clarify: what if the syntax of types were extended. And the outcome printer.
One difficulty is that `React.compnent` is a pre-existing type, and its argument is locally, and invisibly, defined, nominal type.

cristianoc:
In fact, anything that hover shows must be copy-pastable. So it follows that any form of clever hovering needs a syntax extension.

mattdamon108:
> but does the type definition generated by the ppx have dummy location or actual location info?

`type props` which has generated by ppx has dummy loc.

mattdamon108:
I have a question, because I'm not deeply understanding how the extension and compiler works to show the hover. Is it different to show the type information by extension or maybe compiler between the object and record? I think there is difference between V3 and V4.

mattdamon108:
Can we merge and start a new PR for fine tunning? or we still can start a new PR which has base branch to this PR. The thread of this PR gets quite long to show at a glance without unfolding.

cristianoc:
I think we can merge next.
First a question: have you verified that without changes to JSX settings the result of compilation of existing projects is identical?

We want to be reasonably confident that there are no breaking changes.

tests/ppx/react/expected/commentAtTop.res.txt
```diff
@@ -1,5 +1,6 @@
 type props<'msg> = {key?: string, msg: 'msg} // test React JSX file
 
+@react.component
```
cristianoc, line 3: Something to test here is that this works well e.g. with generating interface files with the vscode extension.

cli/reactjs_jsx_ppx.ml
```diff
@@ -2459,12 +2443,20 @@ module V4 = struct
...
             when isOptional name || isLabelled name ->
             getPropTypes ((name, ptyp_loc, type_) :: types) rest
-          | Ptyp_arrow (Nolabel, _type, rest) -> getPropTypes types rest
+          | Ptyp_arrow
+              ( Nolabel,
+                {ptyp_desc = Ptyp_constr ({txt = Lident "unit"}, _)},
+                rest ) ->
+            getPropTypes types rest
+          | Ptyp_arrow (Nolabel, _type, rest) ->
+            hasForwardRef := true;
```
cristianoc, line 2458: Not sure I understand this.
How does one know that there is a forwardRef just because there's a function with a non-labelled arg which is not unit?

Is this check applying on the user code, or on code generated by the ppx?

cristianoc:
OK made a few additional comments/questions.
Next I'll test this a bit.

cristianoc:
Left a comment on https://github.com/rescript-lang/rescript-compiler/pull/5484 to rebase. I'll proceed with testing after that.

cli/reactjs_jsx_ppx.ml
mattdamon108: Good catch. And it is quite a tricky part, because of the usage of forwardRef.
Basically, we are not allowing the nolabel arg in `@react.component`, right? There is one exception to this rule, forwardRef. So, for the interface, it has no way to know what the label is for the type signature. I think we need some assumptions and restrictions at some point. If there is a nolabel type signature after labeled args, the only ref can go there. It will be considered as forwardRef case.

cristianoc: When operating on a forwardRef, do we know that it's a forwardRef to begin with?
If that code was gated by a check "we are inside a forwardRef" then I would have no objection to it.
As it would be clear it's just a simple heuristic, and we don't care because it only applies to forwardRef.

cristianoc: So I imagine a reference `insideForwardRef` which is set. And checked at this point in the code.

mattdamon108:
> First a question: have you verified that without changes to JSX settings the result of compilation of existing projects is identical?
> 
> We want to be reasonably confident that there are no breaking changes.

I've ran the compiler.alpah1 + V4 on my company repo, it has no diff in the generated js files! Nice. 🤞

mattdamon108:
EDIT: I was wrong. It was backward compatibility check with compiler.alpha1 + V3. 😓 

There are some errors with compiler.alpha1 + V4. I'm going to leave it as an issue.

mattdamon108:
Here is the test bed https://github.com/green-labs/sinsunhi-frontend-mirror/tree/test-v4

cristianoc:
Want to close this and continue on https://github.com/rescript-lang/syntax/pull/614 ?

cristianoc:
So backwards compatibility seems OK right?
The only question is debugging V4, correct? That's fine and expected.

mattdamon108:
> So backwards compatibility seems OK right?

Yes, no diff and same output.

> The only question is debugging V4, correct? That's fine and expected.

Yes. I'm back now, gonna leave an issue.

mattdamon108:
> Want to close this and continue on https://github.com/rescript-lang/syntax/pull/614 ?

Good! I'm gonna close this PR.

PR #590, opened 2022-06-28
Callbacks without exponential explosion.

cristianoc:
Follows https://github.com/rescript-lang/syntax/pull/576

### DIFF
--- a/tests/printer/expr/expected/binary.res.txt
+++ b/tests/printer/expr/expected/binary.res.txt
@@ -472,22 +472,18 @@ foo :=
   | _ => 1
   }
 
-<div>
-  {possibleGradeValues
-  |> List.filter(g => g < state.maxGrade)
-  |> List.map(possibleGradeValue =>
+<div> {possibleGradeValues |> List.filter(g =>
+    g < state.maxGrade
+  ) |> List.map(possibleGradeValue =>
     <option key={possibleGradeValue |> string_of_int} value={possibleGradeValue |> string_of_int}>
       {possibleGradeValue |> string_of_int |> str}
     </option>
-  )
-  |> Array.of_list
-  |> ReasonReact.array}
-</div>
+  ) |> Array.of_list |> ReasonReact.array} </div>
 
 let aggregateTotal = (forecast, ~audienceType) =>
-  Js.Nullable.toOption(forecast["audiences"])
-  ->Option.flatMap(item => Js.Dict.get(item, audienceType))
-  ->Option.map(item => {
+  Js.Nullable.toOption(forecast["audiences"])->Option.flatMap(item =>
+    Js.Dict.get(item, audienceType)
+  )->Option.map(item => {
     pages: item["reach"]["pages"],
     views: item["reach"]["views"],
     sample: item["reach"]["sample"],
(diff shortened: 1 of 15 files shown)


tests/printer/expr/expected/let.res.txt
```diff
@@ -47,11 +47,7 @@ let f = x => {
     ->Option.flatMap(Js.Json.decodeString)
     ->Option.map(Js.Date.fromString)
 
-  let b =
-    x
-    ->Js.Dict.get("wm-property")
-    ->Option.flatMap(Js.Json.decodeString)
-    ->Option.flatMap(x =>
+  let b = x->Js.Dict.get("wm-property")->Option.flatMap(Js.Json.decodeString)->Option.flatMap(x =>
```
ryyppy, line 50: AFAIKT this would be a big regression. The community loves tidy pipe chains on multiple lines after a certain line width has been reached.

But I guess this causes most of the perf issues?

cristianoc: There's no causality.
It's an unintended effect of how the exponential explosion is avoided.

tests/printer/expr/expected/binary.res.txt
```diff
@@ -496,16 +492,13 @@ let aggregateTotal = (forecast, ~audienceType) =>
...
-    ->Promise.Js.catch(_ => {Promise.resolved(None)})
-    ->Promise.get(newBalance => {
-      dispatch(
-        LoadAddress(
-          account,
-          newBalance->Belt.Option.flatMap(balance => Eth.make(balance.toString(.))),
-        ),
-      )
-    })
+    library.getBalance(. account)->Promise.Js.catch(_ => {Promise.resolved(None)})->Promise.get(
```
ryyppy, line 495: Promises really profit from per-line pipe formatting. But I guess this might not even be relevant anymore for the "next big feature"?

tests/printer/expr/expected/binary.res.txt
```diff
@@ -472,22 +472,18 @@ foo :=
   | _ => 1
   }
 
-<div>
-  {possibleGradeValues
-  |> List.filter(g => g < state.maxGrade)
-  |> List.map(possibleGradeValue =>
+<div> {possibleGradeValues |> List.filter(g =>
```
IwanKaramazow, line 475: This code seems to be printed in a pretty hard to read way.
How hard is it to preserve the previous format?

tests/printer/expr/expected/binary.res.txt
IwanKaramazow: This seems indeed like a "regression" compared to the previous formatted code

tests/printer/expr/expected/binary.res.txt
cristianoc: No idea! But I have a different approach in mind that should preserve pretty much all the formatting.

tests/printer/expr/expected/callback.res.txt
```diff
@@ -60,10 +60,10 @@ let _ = {
   }
 }
 
-let trees =
-  possibilities->Belt.Array.mapU((. combination) =>
-    combination->Belt.Array.reduceU(Nil, (. tree, curr) => tree->insert(curr))
-  )
+let trees = possibilities->Belt.Array.mapU((. combination) => combination->Belt.Array.reduceU(
```
IwanKaramazow, line 63: Can we keep the formatting on two lines?

tests/printer/expr/expected/binary.res.txt
cristianoc: The fact that it's unclear why the examples have changed, is more worrying than the changes themselves.

tests/printer/expr/expected/callback.res.txt
```diff
@@ -82,11 +82,10 @@ let add2 = (y: coll, e: key) =>
   }
 
 let add2 = (y: coll, e: key) =>
-  if (
-    possibilities->Belt.Array.mapU((. combination) =>
-      combination->Belt.Array.reduceU(Nil, (. tree, curr) => tree->insert(curr))
-    )
-  ) {
+  if possibilities->Belt.Array.mapU((. combination) => combination->Belt.Array.reduceU(Nil, (.
```
IwanKaramazow, line 85: Should this break?

tests/printer/expr/expected/callback.res.txt
```diff
@@ -180,17 +179,13 @@ foo(list => list())
 foo(\"switch" => \"switch"())
 
 // the [] of the array should break
-[
-  fn(x => {
+[fn(x => {
```
IwanKaramazow, line 182: I kinda prefer the previous formatting.

PR #591, opened 2022-06-28
Quick fix to exponential blowup.

cristianoc:
Stop it after a nested threshold.

This has identical output on tests, and solved the perf issue.


leeor:
Sweet! Thank you :-)

cristianoc:
> Sweet! Thank you :-)

Thank you for digging deep into all the issues.

PR #593, opened 2022-07-01
Assertion of parseJsImport fails

Sehun0819:
Assertion in `parseJsImport` function can fail, so I guess it should be properly treated instead of assertion.

- Representative example:
  ```ocaml
  import@
  ```

The problem is, `parseJsFfiDeclaration` happens to return `None` which violates assertion.
I patched it by handling `None` case with an empty declaration, but I'm not sure it is a right way.


src/res_core.ml
```diff
@@ -5411,7 +5411,7 @@ and parseJsImport ~startPos ~attrs p =
       let decl =
         match parseJsFfiDeclaration p with
         | Some decl -> decl
-        | None -> assert false
+        | None -> JsFfi.emptyDecl p.startPos
```
cristianoc, line 5414: See how other errors are handled.
E.g. just a few lines above here:
```ml
Parser.err ~startPos:attrLoc.loc_start ~endPos:attrLoc.loc_end p
        (Diagnostics.message (ErrorMessages.attributeWithoutNode attr));
```

Currently it goes all the way to the end of the file silently (e.g. in the provided test).

cristianoc: Also see this example for where to define the default value to use (inside `Recover`:
```ml
    Parser.err p (Diagnostics.unexpected token p.breadcrumbs);
    Recover.defaultModuleExpr ()

```

Sehun0819: Please check what I added for error handling.
And, do you mean it's better to define the default value(`JsFfi.emptyDecl`, I made) inside `Recover` module?
It seems that all default values in `Recover` are Ast of Ocaml compiler, while Js ffi decls are not.

Sehun0819:
@cristianoc  Please check what I added for error handling.
And, do you mean it's better to define the default value(JsFfi.emptyDecl, I made) inside Recover module?
It seems that all default values in Recover are Ast of Ocaml compiler, while Js ffi decls are not.

IwanKaramazow:
FWIW `import` was a leftover from an experiment that we were never able to ship. Seems like the community found it a very bad thing (https://github.com/rescript-lang/syntax/pull/217). Maybe the code path in the parser should just be removed?

Sehun0819:
@IwanKaramazow Thank you for informing of what was going on.
I had been curious because the source code seemed to accept more grammar than documentation, finally got an answer :)

cristianoc:
Let's remove it.

cristianoc:
I guess it means starting over with this PR

Sehun0819:
Then, should I close this PR? 🤔

cristianoc:
Or just undo the changes. Don't think anything will stay.

cristianoc:
The idea is to literally remove parsing "import".

cristianoc:
The same goes for "export".

cristianoc:
Here's what it would look like: https://github.com/rescript-lang/syntax/pull/597

after removing all the resulting dead code reported from the analysis which runs on `make test`.

The same goes for `export`.

Issue #594, opened 2022-07-01
Making parser consistent with #542


Sehun0819:
Follow-up steps for #542 should be taken but I have no idea how to do.

- Motivating example(I’m guessing there would be  a few more, I’ve not checked though):  `#`

Above simple code crashes the parser because `parseIdent` calls `Parser.next` with `Eof`, eventually violates an assertion of `res_parser.ml` added by #542. I thought it could be fixed easily by adding another pattern matching for `Eof` or substitute `Parser.next` with `Parser.nextUnsafe`(as far as it does not results another infinite loop), but neither pass the static analyzer. Any solution?

https://github.com/rescript-lang/syntax/blob/08029afafeef01cda2366c580b0b06d376b31d54/src/res_core.ml#L564-L586

Sehun0819:
@cristianoc

cristianoc:
How about checking for `Eof` before calling `parseIndent`.

Sehun0819:
Please check #596.
Seems to work, though a few code repeat was needed.

cristianoc:
Looks good.
Will take a closer look later.

Sehun0819:
Thank you.
Please let me know after taking a look later, as there are other similar cases to be treated.

PR #596, opened 2022-07-02
checking Eof before calling parseIdent

Sehun0819:
Handling #594.
Guards are added before `parseIdent`, to prevent it from calling `Parser.next` on `Eof`.

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -600,7 +600,11 @@ let parseHashIdent ~startPos p =
     in
     Parser.next p;
     (i, mkLoc startPos p.prevEndPos)
-  | _ -> parseIdent ~startPos ~msg:ErrorMessages.variantIdent p
+  | token ->
+    if token = Eof then (
+      Parser.err ~startPos p (Diagnostics.unexpected token p.breadcrumbs);
+      ("", mkLoc startPos p.prevEndPos))
+    else parseIdent ~startPos ~msg:ErrorMessages.variantIdent p
 
 (* Ldot (Ldot (Lident "Foo", "Bar"), "baz") *)
 let parseValuePath p =
@@ -1090,7 +1094,12 @@ let rec parsePattern ?(alias = true) ?(or_ = true) p =
             in
             Parser.next p;
             (i, mkLoc startPos p.prevEndPos)
-          | _ -> parseIdent ~msg:ErrorMessages.variantIdent ~startPos p
+          | token ->
+            if token = Eof then (
+              Parser.err ~startPos p
+                (Diagnostics.unexpected token p.breadcrumbs);
+              ("", mkLoc startPos p.prevEndPos))
+            else parseIdent ~msg:ErrorMessages.variantIdent ~startPos p
         in
         match p.Parser.token with
         | Lparen -> parseVariantPatternArgs p ident startPos attrs
@@ -3808,7 +3817,11 @@ and parseAtomicTypExpr ~attrs p =
     | SingleQuote ->
       Parser.next p;
       let ident, loc =
-        parseIdent ~msg:ErrorMessages.typeVar ~startPos:p.startPos p
+        if p.Parser.token = Eof then (
+          Parser.err ~startPos:p.startPos p
+            (Diagnostics.unexpected p.Parser.token p.breadcrumbs);
+          ("", mkLoc p.startPos p.prevEndPos))
+        else parseIdent ~msg:ErrorMessages.typeVar ~startPos:p.startPos p
       in
       Ast_helper.Typ.var ~loc ~attrs ident
     | Underscore ->
@@ -4596,7 +4609,11 @@ and parseTypeParam p =
   | SingleQuote ->
     Parser.next p;
     let ident, loc =
-      parseIdent ~msg:ErrorMessages.typeParam ~startPos:p.startPos p
+      if p.Parser.token = Eof then (
+        Parser.err ~startPos:p.startPos p
+          (Diagnostics.unexpected p.Parser.token p.breadcrumbs);
+        ("", mkLoc p.startPos p.prevEndPos))
+      else parseIdent ~msg:ErrorMessages.typeParam ~startPos:p.startPos p
     in
     Some (Ast_helper.Typ.var ~loc ident, variance)
   | Underscore ->


cristianoc:
Looking good.
Code duplication is fine as it's very very little compared to having parser.next return an option and having to check every single one.

A helper function for the error message should be more then enough to reduce boilerplate.

Sehun0819:
Thank you.
At first, I was thinking of fixing only apparent bugs(that is, found by fuzzer), so that innocent codes aren't affected. But it seems like more conservative(or aggressive) fixing is needed because there are so many `Parser.next` in `res_core.ml`.

https://github.com/rescript-lang/syntax/blob/08029afafeef01cda2366c580b0b06d376b31d54/src/res_core.ml#L3603-L3617

Concretely, `Parser.next` in L3604 would be safe, while one in L3610 seems highly suspicious though I've seen no concrete example yet. Would you give me an opinion?

cristianoc:
There's no harm in doing all the non-obvious cases

For comparison there's a PR showing how horrible the code would look using exception analysis

cristianoc:
This PR: https://github.com/rescript-lang/syntax/pull/543
which has over 1000 code changes, is what we're trying to avoid here.

src/res_core.ml
```diff
@@ -1906,7 +1915,7 @@ and parseFirstClassModuleExpr ~startPos p =
 and parseBracketAccess p expr startPos =
   Parser.leaveBreadcrumb p Grammar.ExprArrayAccess;
   let lbracket = p.startPos in
-  Parser.next p;
+  Parser.expect Lbracket p;
```
Sehun0819, line 1918: Though just using `Parser.next` is safe as `p` is always `Lbracket`, this change would make it more clear.

Sehun0819:
I looked over every `Parser.next` usages in `res_core.ml`.
- Because it is hard to fix `skipTokensAndMaybeRetry` directly(like `parseIdent`),  I added guards right before call sites.
- When it seems ok, I substituted `Parser.next` with `Parser.nextUnsafe`.

src/res_core.ml
```diff
@@ -1115,11 +1128,14 @@ let rec parsePattern ?(alias = true) ?(or_ = true) p =
       Ast_helper.Pat.extension ~loc ~attrs extension
     | token -> (
       Parser.err p (Diagnostics.unexpected token p.breadcrumbs);
-      match
-        skipTokensAndMaybeRetry p ~isStartOfGrammar:Grammar.isAtomicPatternStart
-      with
-      | None -> Recover.defaultPattern ()
-      | Some () -> parsePattern p)
+      if p.Parser.token = Eof then Recover.defaultPattern ()
```
cristianoc, line 1131: Error message missing

src/res_core.ml
```diff
@@ -556,10 +556,14 @@ let rec parseLident p =
     Parser.next p;
     let loc = mkLoc startPos p.prevEndPos in
     (ident, loc)
-  | _ -> (
-    match recoverLident p with
-    | Some () -> parseLident p
-    | None -> ("_", mkLoc startPos p.prevEndPos))
+  | token -> (
+    if token = Eof then (
```
cristianoc, line 560: You can move this to a case in pattern matching

src/res_core.ml
```diff
@@ -600,7 +604,11 @@ let parseHashIdent ~startPos p =
     in
     Parser.next p;
     (i, mkLoc startPos p.prevEndPos)
-  | _ -> parseIdent ~startPos ~msg:ErrorMessages.variantIdent p
+  | token ->
+    if token = Eof then (
```
cristianoc, line 608: You can move this to a case in the pattern matching.

src/res_core.ml
```diff
@@ -1090,7 +1098,12 @@ let rec parsePattern ?(alias = true) ?(or_ = true) p =
             in
             Parser.next p;
             (i, mkLoc startPos p.prevEndPos)
-          | _ -> parseIdent ~msg:ErrorMessages.variantIdent ~startPos p
+          | token ->
+            if token = Eof then (
```
cristianoc, line 1102: Case in pattern matching

src/res_core.ml
cristianoc: Can be moved to case in pattern matching

src/res_core.ml
```diff
@@ -1862,11 +1878,13 @@ and parseAtomicExpr p =
     | token -> (
       let errPos = p.prevEndPos in
       Parser.err ~startPos:errPos p (Diagnostics.unexpected token p.breadcrumbs);
-      match
-        skipTokensAndMaybeRetry p ~isStartOfGrammar:Grammar.isAtomicExprStart
-      with
-      | None -> Recover.defaultExpr ()
-      | Some () -> parseAtomicExpr p)
+      if p.Parser.token = Eof then Recover.defaultExpr ()
```
cristianoc, line 1881: Error missing

src/res_core.ml
```diff
@@ -3856,14 +3881,17 @@ and parseAtomicTypExpr ~attrs p =
...
       Parser.err p (Diagnostics.unexpected token p.breadcrumbs);
-      match
-        skipTokensAndMaybeRetry p ~isStartOfGrammar:Grammar.isAtomicTypExprStart
-      with
-      | Some () -> parseAtomicTypExpr ~attrs p
-      | None ->
-        Parser.err ~startPos:p.prevEndPos p
-          (Diagnostics.unexpected token p.breadcrumbs);
-        Recover.defaultType ())
+      if p.Parser.token = Eof then Recover.defaultType ()
```
cristianoc, line 3884: Error missing

Sehun0819:
How about now?

src/res_core.ml
```diff
@@ -556,14 +556,13 @@ let rec parseLident p =
...
     (ident, loc)
-  | token -> (
-    if token = Eof then (
-      Parser.err ~startPos p (Diagnostics.unexpected token p.breadcrumbs);
-      ("_", mkLoc startPos p.prevEndPos))
-    else
-      match recoverLident p with
-      | Some () -> parseLident p
-      | None -> ("_", mkLoc startPos p.prevEndPos))
+  | token when token = Eof ->
```
cristianoc, line 559: simply `| Eof`

cristianoc:
Merged as needed for release

Issue #365, opened 2021-04-02
String interpolation within an empty string is a no-op


johnridesabike:
```res
let x = `${[1,2]}`
```
Output:
```js
var x = [1, 2];
```

Maybe this isn't a bug? I would expect it to cause a type error though.

cknitt:
@ryyppy @IwanKaramazow This seems to be fixed already, so we can close this issue.

When I build rescript with the latest master of the syntax repo, I get

```
  We've found a bug for you!
  /Users/christoph/projects/playground/rescript-project-template/src/Demo.res:1:12-17

  1 │ let x = `${[1, 2]}`
  2 │

  This has type: array<'a>
  Somewhere wanted: string
```

PR #602, opened 2022-07-06
Do not generate leading and trailing `""` in template expressions.

cristianoc:
Fixes https://github.com/rescript-lang/rescript-compiler/issues/5521

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -2158,8 +2158,10 @@ and parseTemplateExpr ?(prefix = "js") p =
         Ast_helper.Exp.constant ~attrs:[templateLiteralAttr] ~loc
           (Pconst_string (txt, Some prefix))
       in
-      Ast_helper.Exp.apply ~attrs:[templateLiteralAttr] ~loc hiddenOperator
-        [(Nolabel, acc); (Nolabel, str)]
+      if txt = "" then acc
+      else
+        Ast_helper.Exp.apply ~attrs:[templateLiteralAttr] ~loc hiddenOperator
+          [(Nolabel, acc); (Nolabel, str)]
     | TemplatePart txt ->
       Parser.next p;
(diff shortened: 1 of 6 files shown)


cknitt:
Is this not reintroducing #365?

cristianoc:
> Is this not reintroducing #365?

Good catch. And it has other problems too.

cristianoc:
Just writing down some notes here.

This reliance on indirect type checking by putting `""` at either end of string concatenation seems brittle.
Also there seem to be several layers of sediment left presumably from old ways of doing things.

So ``` `stuff` ``` is the same as ``` js`stuff` ``` by convention I think.

Then there is ``` j`stuff` ``` which somehow obeys different rules. So in ``` j`$x` ``` the `$x` actually has a meaning. And `x` does not need to be a string. The consequence of this is that `"" + x` is not the same as `x` when `x` is not a string. So removing empty string concatenation in the back-end of the compiler is also delicate as it's easy to do it wrong (`3 + ""` can't be removed).

Then there's ``` json`stuff` ``` which I don't know maybe it's the same as `j` but old, not really sure. Are they really treated in the same way at every stage in the compiler? Not sure.

All this is represented internally by putting together strings that have a tag "j" or "js" or "json".
In addition to all that, strings produced by the parser are now by default unicode, and that uses the tag "*j". But, there's also the OCaml parser for `.ml` files which will never generate "*j" for normal strings.

In addition to all this, there's some half attempt to also use a type "unicode" inside the back-end of the compiler, which seems incomplete.

Also, there's a quoting mechanism that happens on dump (code generation) which depends on which kind of string it is.

Goes without saying, all this needs a good cleanup.

cristianoc:
Moving these thoughts to an issue in the compiler repo https://github.com/rescript-lang/rescript-compiler/issues/5522

PR #603, opened 2022-07-07
Make confix flexible

cristianoc:


mattdamon108:
@cristianoc Can I ask your intention to sync? Is it what you're implementing?
```rescript
// A.res
@@jsxConfig({version: 3})

@react.compoent
let make = () => body // 1

@@jsxConfig({version: 4})

@react.compoent
let make = () => body // 2
```
```rescript
// B.res

@react.component
let make = () => body // 3
```

1. A.res, 1 -> V3
2. A.res, 2 -> V4
3. B.res, 3 -> by bsconfig.json

If it is your intention, then I have an idea without merging V3 and V4 into one. I prefer to separate it for the future maintaining.

cristianoc:
> > Yes. Feel free to continue from where I have left.
> > As mentioned, there's really no fusion between V3 and V4. And it's easy to copy back to separate files later.
> 
> Ah! Finally, I figure your intention out. Thank you for your implementation, I'll look into it. I have an idea to implement your intention, I'll ping you once I've done it. Please check that out later. We can compare and mature the trade-off between implementations, then choose it.

I have cleaned up a bunch of things.
I think if you continue from this PR it's better.
As we avoid tens of review back and forth later to clean things up.
The structure of the code has not changed really.

cristianoc:
I assume your idea is not incompatible with the cleanup I have done.

mattdamon108:
I like what you've done in this PR. But, I have a little idea to implement your intention from my PR, I'll show you how it looks. Maybe if you don't like it, then it will be totally fine with me. We can keep developing the V4 from this PR. I think that showing the codes would be the best communication for this implementation.

cristianoc:
> I like what you've done in this PR. But, I have a little idea to implement your intention from my PR, I'll show you how it looks. Maybe if you don't like it, then it will be totally fine with me. We can keep developing the V4 from this PR. I think that showing the codes would be the best communication for this implementation.

Sure go for it.
Though merging to this later might take more time. Or maybe not. Let's see.

cristianoc:
I guess your idea is you'll try to do 2 separate traversals, one for V3 and one for V4, and each one will transform only the things that are in V3/V4 current config. If that's the idea, let's see how it comes.

cristianoc:
That won't work for inner modules.

mattdamon108:
I just made a commit of it https://github.com/rescript-lang/syntax/pull/586/commits/572a8c7dbfb9e0fc1251ea1e452363263cb89f64

cristianoc:
> I just made a commit of it [572a8c7](https://github.com/rescript-lang/syntax/commit/572a8c7dbfb9e0fc1251ea1e452363263cb89f64)

Yea it does not work for inner modules.

mattdamon108:
I think there is a test case of inner modules. Is it what you refer to? https://github.com/rescript-lang/syntax/blob/master/tests/ppx/react/innerModule.res

cristianoc:
> I think there is a test case of inner modules. Is it what you refer to? https://github.com/rescript-lang/syntax/blob/master/tests/ppx/react/innerModule.res

I mean switch between 3 and 4 inside an inner module.

mattdamon108:
> I mean switch between 3 and 4 inside an inner module.

Wow, I didn't notice that `@@jsxConfig` can be used inside modules not top-level only.
```rescript
@@jsxConfig({version: 4})

@react.component
let make = () => {
  @@jsxConfig({version: 3})
  <div> ... </div>
}
```
Is it what you mean switching inside the inner module?

mattdamon108:
Correct, my idea is not working for the inner module.

cristianoc:
> I think I didn't fully understand the spec of `@@jsxConfig`.

There are many details that can be easily missed. That's why I wanted to code directly to show the idea.

mattdamon108:
Agree!
Alright, we can move forward from this PR. I'm afraid I can't push directly this branch, can I?

cristianoc:
> Agree!
> 
> Alright, we can move forward from this PR. I'm afraid I can't push directly this branch, can I?

You can just copy the commits. Or I'll move them later to your branch when I'm back.

mattdamon108:
> You can just copy the commits. Or I'll move them later to your branch when I'm back.

I'll cherry-pick your commits!

cristianoc:
Incorporated already.

PR #606, opened 2022-07-09
Reenable and fix deprecation warnings

cknitt:
Motivation: I am trying to get rid of the warnings when building the ReScript compiler, and some of them are coming from the syntax module.

* Reenabled deprecation warnings.
* Removed `Pervasives` module from `compiler-libs-406`.
* Fixed deprecation warnings related to `Pervasives`.
* Fixed deprecation warnings related to `pp_set_formatter_tag_functions`.

Tested: Still builds fine with OCaml 4.10.


cristianoc:
This look great.
Would you do the same for the compiler, and rescript-vscode, check that all still works.
Then we can proceed and do it with all 3.

cknitt:
I was trying to do it step by step. I have already checked that the compiler still builds with this version of the syntax repo. I now also checked that the compiler tests are still running fine.

Will create a similar PR for the rescript-vscode repo. 👍

cristianoc (approved):

PR #610, opened 2022-07-15
Fix comments after function parameters disappearing

cknitt:
This PR fixes c1 and c2 in the test disappearing. (c3 and c4 print the same as they would have before, had they been in the test.)

I was looking into this some time ago and did not submit the PR as I wasn't happy with the locations where the comments are printed. But I think for now it is good enough that they don't disappear anymore, the location can be improved separately.

### DIFF
--- a/src/res_printer.ml
+++ b/src/res_printer.ml
@@ -4585,11 +4585,12 @@ and printExprFunParameters ~customLayout ~inCallback ~uncurried ~hasConstraint
        attrs = [];
        lbl = Asttypes.Nolabel;
        defaultExpr = None;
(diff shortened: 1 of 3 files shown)


cristianoc (approved):
As a general rule, how does one know that there is no double printing?
Are comments somehow consumed, or does one need to rely on some implicit global property of the entire printer?

cknitt:
They are actually consumed (had to look that up, too, now), see e.g. https://github.com/rescript-lang/syntax/blob/241fefd8e18718fa6080be3a70764b79f2fad5c1/src/res_printer.ml#L280

PR #614, opened 2022-07-19
Jsx ppx v4 part2

cristianoc:
This moves https://github.com/rescript-lang/syntax/pull/586 to a branch on the repo.


cristianoc:
OK so I've tested this with the genType example project for TypeScript.
With V3 no change at all. Also with `"version": 3` in the config it's fine.

With V4 I get:
```
FAILED: src/ImportHookDefault.cmj

  We've found a bug for you!
  /Users/cristianocalcagno/GitHub/genType/examples/typescript-react-example/src/ImportHookDefault.res

  Multiple definition of the type name props.
  Names must be unique in a given structure or signature.
```
which is not surprising.

cristianoc:
In terms of merging this, there's still the outstanding comment about forwardRef.
The rest seems fine, and could continue in another PR.
Let me know if you have questions about forwardRef.

cristianoc:
> With V4 I get:
> 
> ```
> FAILED: src/ImportHookDefault.cmj
> 
>   We've found a bug for you!
>   /Users/cristianocalcagno/GitHub/genType/examples/typescript-react-example/src/ImportHookDefault.res
> 
>   Multiple definition of the type name props.
>   Names must be unique in a given structure or signature.
> ```
> 
> which is not surprising.

While this is a breaking change, I'm wondering whether it could be turned into a net positive.
If it make sense to enforce: at most one component for each module.
mmm that might be getting in the way of libraries or component factories etc

worth thinking about anyway

cristianoc:
One possible view is that for component definitions, what the ppx does is completely transparent and documented.
So one can just write a component definition directly without even using the ppx.
That is at least an escape hatch that one could use.

So this is about just accepting the current situation and putting it in context.

A separate thing to think about is to tweak the name of the props type being defined.

mattdamon108:
> One possible view is that for component definitions, what the ppx does is completely transparent and documented. So one can just write a component definition directly without even using the ppx. That is at least an escape hatch that one could use.
> 
> So this is about just accepting the current situation and putting it in context.
> 
> A separate thing to think about is to tweak the name of the props type being defined.

I had noticed it with the test case in `innerModule` before, I think it is worth thinking about: one component per module. If we tweak the name of the props type, the error will be gone in the definition site. In the caller site, the uppercase, the application site, is parsed only to `Foo.make` anyway. So, IMHO, I agree with your opinion that accepting the current situation and putting it in the context as you said.

cristianoc:
Not sure I understand which one you're proposing:
1 rename "prop" where necessary to avoid multiple definitions 
2 enforce one component per module (so many inner modules are ok)

Which one are you thinking about?

mattdamon108:
> 2 enforce one component per module (so many inner modules are ok)

This is what I'm proposing.

cristianoc:
> > 2 enforce one component per module (so many inner modules are ok)
> 
> 
> 
> This is what I'm proposing.

I think this needs to be enforced directly in the ppx. Not by a compiler error.

cristianoc:
So for each module in the ast, check that at most 1 component definition is there.

mattdamon108:
Can I ask what your proposal is?

> I think this needs to be enforced directly in the ppx. Not by a compiler error.

I guess it is not a difficult job to achieve with `ast_iterator`.

cristianoc:
> Can I ask what your proposal is?
> 
> > I think this needs to be enforced directly in the ppx. Not by a compiler error.
> 
> I guess it is not a difficult job to achieve with `ast_iterator`.
> 
> EDIT: not even `ast_iterator` is needed, I guess. We can identify how many react components are there at the module level.

Exactly. It fits naturally in the existing ppx mapper.

cristianoc (commented):
There are a lot of visual changes so don't know exactly everything that has changed.
But notice that `hasReactComponent` could be changed in a sub-module, and should be restore to its previous value when processing the submodule is finished.

cli/reactjs_jsx_ppx.ml
```diff
@@ -2689,6 +2697,7 @@ let getMapper ~config =
   in
   let signature mapper items =
     let oldConfig = saveConfig () in
+    let hasReactComponent = ref false in
```
cristianoc, line 2702: If you put this in config, then you get it restored for free after sub-modules are done.

mattdamon108:
> There are a lot of visual changes so don't know exactly everything that has changed. But notice that `hasReactComponent` could be changed in a sub-module, and should be restore to its previous value when processing the submodule is finished.

Most diff is cased by indent by if else expression added. Actually, `hasReactComponent` value is restored per modules because it is assigned inside the structure mapper.

cli/reactjs_jsx_ppx.ml
mattdamon108: Yes, it will work. I thought that too, but I little hesitated because the config means the config of jsx. It seems little unmatched with `hasReactComponent`, doesn't it?

mattdamon108: hmm.. There is `nestedModules` inside the config already. I think `hasReactComponent` could be there too. I'll fix it.

cristianoc: Yes `nestedModules` is in there for exactly this reason: avoid passing ad-hoc state when it can be all together.
If you think the type should have another better name, go ahead renaming it.

mattdamon108: I fixed it by adding `hasReactComponent` into the config. I realized why I felt a little weird about it.
It is because the config keeps the values until it hits another config update, even though it is restored when it goes to the upper level, but it doesn't set default when it goes to the deeper level. Therefore config.hasReactComponent should be set default(false) in the structure and signature mapper.
I think `hasReactComponent` doesn't fit in the config.

cristianoc: This is not very important. You could put in inside or leave it outside.

cristianoc: `module_binding` is only for implementations.
I've quickly edited so it supports interfaces as well.

mattdamon108: Ah! Good catch. Thanks!

mattdamon108:
Shouldn't config.hasReactComponent be restored?

mattdamon108:
I add to restore the old value of config.hasReactComponent

cristianoc:
This is looking good for an initial merge as WIP. Seems low-risk as existing projects are using V3 should not be affected.

PR #618, opened 2022-07-21
Fix optional prop value example

mattdamon108:
This PR is to investigate how to prevent the breaking change for users regarding the usage of optional prop values between jsx ppx V3 -> V4.


mattdamon108:
I'm going to close this PR and seek the solution in the ongoing PR #617

Issue #627, opened 2022-08-24
Parse error for optional field with attribute


cknitt:
This

```rescript
type x = {
  @someAttr
  a?: int,
}
```

gives

```
  Syntax error!
  Test.res:3:4

  1 │ type x = {
  2 │   @someAttr
  3 │   a?: int,
  4 │ }
  5 │

  I'm not sure what to parse here when looking at "?".
```

cristianoc:
Thank you for the report.
What was the original example like? Contrived or everyday?

cknitt:
Originally I was thinking about whether we still need `@deriving(abstract)` with `@optional` on a field or if could deprecate that in favor of records with optional fields.

Then I tried out what would happen if I had an optional field + the deriving abstract `@optional` annotation and came across the above problem.

However, one could also run into this when trying to use `@as(...)` or maybe some attribute for JSON serialization like what decco uses.

cristianoc:
Thanks. Marking for 10.1.

cristianoc:
So this seems to affect only single-field records: https://github.com/rescript-lang/syntax/pull/628

PR #628, opened 2022-08-26
Issue with single optional field with attributes.

cristianoc:
Add test. Multiple fields works fine.
See: https://github.com/rescript-lang/syntax/issues/627

### DIFF
--- a/src/res_core.ml
+++ b/src/res_core.ml
@@ -4286,6 +4286,7 @@ and parseFieldDeclaration p =
     match p.token with
     | _ -> parseLident p
   in
+  let optional = parseOptionalLabel p in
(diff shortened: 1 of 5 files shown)


cristianoc:
Probably something around this change: https://github.com/rescript-lang/syntax/pull/588/files#diff-ae497450e5424068bdb6808050ebf751eccbc481ca82c61c2d0067ba1636e0b0R4291

The flow is not so simple for single-field records as they are recognised quite late in the parsing process (great potential for ambiguity).

cristianoc:
@IwanKaramazow can you take a look? Can this be done in a simpler way?

PR #630, opened 2022-09-10
jsx: improve ref prop error message

glennsl:
The [current rescript-react docs on forwarding refs](https://rescript-lang.org/docs/react/latest/forwarding-refs) discourages the use of `forwardRef` in favor of explicitly passing props. It seems a bit inconsistent then for the PPX to only recommend the `forwardRef` API then, if it should be mentioned at all. FOr now I've opted for mentioning both without encouraging one over the other.

The docs themselves could also be improved in a couple of ways:

1. It doesn't mention that the forwarded prop cannot be named `ref`, which along with this error message confused me quite a bit.
2. It's not clear how to specify the type of the component in an interface file when using `forwardRef`.


cristianoc:
Thanks, makes sense.
And indeed the docs can be improved.

PR #633, opened 2022-09-14
Add loc for returned component body

mattdamon108:
This PR fixes the issue https://github.com/rescript-lang/rescript-compiler/issues/5635
It show the error with loc now.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,10 +18,11 @@
 
   - :boom: when V4 is activated, at most one component is allowed for each module.
 
-- Add support for empty record literal `{}` for records with only optional fields, and type definition of empty record (e.g. `type empty = {}`)  https://github.com/rescript-lang/syntax/pull/632
+- Add support for empty record literal `{}` for records with only optional fields, and type definition of empty record (e.g. `type empty = {}`) https://github.com/rescript-lang/syntax/pull/632
 
 #### :bug: Bug Fix
 
+- Fix location issue in error messages with JSX V4 where the body of the component is an application https://github.com/rescript-lang/syntax/pull/633
 - Fix issue where the printer would omit attributes for `->` and `|>` https://github.com/rescript-lang/syntax/pull/629
 - Fix printing of optional fields in records https://github.com/rescript-lang/rescript-compiler/issues/5654
 
--- a/cli/reactjs_jsx_ppx.ml
+++ b/cli/reactjs_jsx_ppx.ml
@@ -2331,7 +2331,7 @@ module V4 = struct
                       (Pat.var ~loc:emptyLoc {loc = emptyLoc; txt})
                       fullExpression;
                   ]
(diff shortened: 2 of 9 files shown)


cristianoc:
@mattdamon108  thanks for spotting the location of the issue.

Now for the error message, looking at the error after PPX.

Desugared:
```res
module Foo = {
  let compare = (nextProps, currentProps) => nextProps["foo"] == currentProps["foo"]
  let memo = React.memoCustomCompareProps(_, compare)
  type props<'foo> = {key?: string, foo: 'foo}

  let make = ({foo, _}: props<string>) => React.string(foo)
  let make = memo({
    let \"Tst$Foo" = (props: props<_>) => make(props)

    \"Tst$Foo"
  })
}
```

Error:
```

  We've found a bug for you!
  /Users/cristianocalcagno/GitHub/rescript-vscode/analysis/tests/src/Tst.res:10:5-14

   8 │     let \"Tst$Foo" = (props: props<_>) => make(props)
   9 │ 
  10 │     \"Tst$Foo".    <-------
  11 │   })
  12 │ }

  This has type: props<string> => React.element
  Somewhere wanted:
    React.component<({.."foo": 'b} as 'a)> (defined as 'a => React.element)
  
  The incompatible parts:
    props<string> vs 'a
```

cristianoc:
Alternative JSX behaviour (don't let-define `Tst$Foo`):
```res
module Foo = {
  let compare = (nextProps, currentProps) => nextProps["foo"] == currentProps["foo"]
  let memo = React.memoCustomCompareProps(_, compare)
  type props<'foo> = {key?: string, foo: 'foo}

  let make = ({foo, _}: props<string>) => React.string(foo)
  let make = memo({
    (props: props<_>) => make(props)
  })
}
```

Error:
Screenshot 2022-09-14 at 09 34 30

```

```

cristianoc:
The second one is clearer, but I'm not sure how clearly it can be expressed w.r.t. the original code.

cristianoc:
@mattdamon108  I'm thinking of merging the PR as it is now.
Thoughts?

mattdamon108:
> Alternative JSX behaviour (don't let-define `Tst$Foo`):

Can you show me the js output? The lef-define such as `Tst$Foo` is a secret recipe to make the js output function name starting with the capital letter.

cristianoc (approved):
Thanks for the fix!
Some discussion in the conversation part of the PR.
If anything, we can tackle that later on.

mattdamon108:
Sorry for missing to add tests 😓 I won't forget next time.

cristianoc:
> > Alternative JSX behaviour (don't let-define `Tst$Foo`):
> 
> Can you show me the js output? The lef-define such as `Tst$Foo` is a secret recipe to make the js output function name starting with the capital letter.

That magic does not apply when the top-level of the body is an application: it simply gets lost.

PR #634, opened 2022-09-16
Use empty record in JSX V4.

cristianoc:


mattdamon108 (approved):
I didn't grab the time to work on this. I'll try to grab the time to remove key field in props type.

cristianoc:
Just doing some little tests here.
Noticed I should upstream the empty record parser to the editor or one gets a syntax error while typing.

PR #635, opened 2022-09-16
Remove key prop from JSX props type

mattdamon108:
* Remove the key prop from the props type
* Add magic ~~`React.addKeyProp`~~`Jsx.addKeyProp` for the type checking


mattdamon108:
I noticed that `type domProps` has the `key: string` too. It is used for the lowercase such as `<div key="k" />`. Is it also needed to be removed? https://github.com/rescript-lang/rescript-react/blob/46944b490ee1923cfb69fa5ef0bd3095f93d18de/src/ReactDOM.res#L58

mattdamon108:
Does the lowercase have the same context? https://github.com/rescript-lang/rescript-compiler/issues/5646#issuecomment-1243607315

mattdamon108:
I think the lowercase wouldn't be a problem, because it is still use the `@deriving(abstract)`.

```rescript
ReactDOMRe.createDOMElementVariadic("div", ~props=ReactDOMRe.domProps(~key="k", ()), []),
```

cristianoc:
I think lowercase is OK

cristianoc:
Would you update the spec too?

mattdamon108:
> Would you update the spec too?

https://github.com/rescript-lang/syntax/pull/635/commits/0fbf9a7abe521f6d34bcd41fd8d6ef5b7aa93b2f

cristianoc (approved):
This is great.
Anything missing?

mattdamon108:
> This is great. Anything missing?

Thanks.
No, it seems done.

PR #638, opened 2022-09-18
Refactor PPX: move across files and clean up exceptions

cristianoc:
Move PPX V3 and V4 back to different files.
Clean up treatment of exceptions.


cli/dune
```diff
@@ -1,7 +1,8 @@
 (executable
  (name res_cli)
  (public_name rescript)
+ (modes byte exe)
```
mattdamon108, line 4: Side comment: `byte` for debugger maybe?
Any runtime debugger suggestion? I haven't tried it yet.

cristianoc: This is because otherwise dune does not generate files required by reanalyze.
But for standalone executable it should be possible.
Notice the debugger is not so user friendly though so one needs to try a bit and asses when it's worth using it.

PR #640, opened 2022-09-18
Make empty props as generic type

mattdamon108:
This PR fixes the inconsistency of props type depending on the count of props.
```rescript
// no props
type props = {}

// props >= 1
type props<'a, 'b, ...> = {a: 'a, b: 'b, ...}
```
Therefore, it causes the type error
```rescript
module A = {
  type props = {}
}

let c = React.createElement(A.make, Jsx.addKeyProp({}: A.props<_>, "key")) // Error
// The type A.props is not generic so expects no arguments,
// but is here applied to 1 argument(s)
```
Simply, dropping the type annotation from the first argument of `Jsx.addKeyProp` is not solving a problem. If it doesn't have the type annotation `A.props<_>`, it causes another case of type error.
```rescript
module A = {
  type props<'x> = {x: 'x}
}

let c = React.createElement(A.make, Jsx.addKeyProp({x: "x"}, "key")) // Error
// record can't find ...
```

The solution is making the props type as generic even it has no props at all.
```rescript
type props<_> = {}
let make = (_: props<_>) => { ... }
```

### DIFF
--- a/cli/reactjs_jsx_v4.ml
+++ b/cli/reactjs_jsx_v4.ml
@@ -35,6 +35,10 @@ let constantString ~loc str =
 (* {} empty record *)
 let emptyRecord ~loc = Exp.record ~loc [] None
 
+(* <_> *)
+let genericCoreTypes ~loc =
+  [{ptyp_desc = Ptyp_any; ptyp_loc = loc; ptyp_attributes = []}]
+
 let safeTypeFromValue valueStr =
   let valueStr = getLabel valueStr in
   if valueStr = "" || (valueStr.[0] [@doesNotRaise]) <> '_' then valueStr
@@ -228,13 +232,18 @@ let recordFromProps ~loc ~removeKey callArguments =
       pexp_attributes = [];
     }
 
-(* make type params for make fn arguments *)
+(* Make type params of the props type *)
 (* let make = ({id, name, children}: props<'id, 'name, 'children>) *)
-let makePropsTypeParamsTvar namedTypeList =
-  namedTypeList
-  |> List.filter_map (fun (_isOptional, label, _, _interiorType) ->
-         if label = "key" then None
-         else Some (Typ.var @@ safeTypeFromValue (Labelled label)))
+let makePropsTypeParamsTvar ~default namedTypeList =
+  let params =
+    namedTypeList
+    |> List.filter_map (fun (_isOptional, label, _, _interiorType) ->
+           if label = "key" then None
+           else Some (Typ.var @@ safeTypeFromValue (Labelled label)))
+  in
+  match params with
+  | [] -> default
+  | _ -> params
 
 let stripOption coreType =
   match coreType with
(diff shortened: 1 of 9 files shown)


mattdamon108:
Haven't noticed the error in case of empty props until I ran the latest master on the larger project 😓

cristianoc:
Perhaps there's another way.

First, the type of `addKeyProp` is not correct: it should return the prop type, so it needs a type constraint
```ml
let addKeyProp (o : 'props) (k : string) : 'props =
  Obj.magic (Js.Obj.assign (Obj.magic o) [%obj { key = k }])
  [@@inline]
```

Second, adding a `createElementWithKey` with the appropriate type seems to take care of the problem: this compiles fine

```res
module C = {
  @react.component
  let make = () => React.null
}

module D = {
  @react.component
  let make = (~x="", ~y="") => React.string(x++y)
}

let createElementWithKey = (comp, props, key) => React.createElement(comp, Jsx.addKeyProp(props, key))

let c1 = React.createElement(C.make, {})
let c2 = createElementWithKey(C.make, {}, "")
let d1 = React.createElement(D.make, {})
let d2 = createElementWithKey(D.make,{}, "")
let d3 = createElementWithKey(D.make, {x:""}, "")
```

So another way to go is to expose e.g. `Jsx. createElementWithKey` of the appropriate type instead of `Jsx.addKeyProp`.
The type being `(React.component<'a>, 'a, string) => React.element`.

mattdamon108:
> Perhaps there's another way.
> 
> First, the type of `addKeyProp` is not correct: it should return the prop type, so it needs a type constraint
> 
> ```ocaml
> let addKeyProp (o : 'props) (k : string) : 'props =
>   Obj.magic (Js.Obj.assign (Obj.magic o) [%obj { key = k }])
>   [@@inline]
> ```
> 
> Second, adding a `createElementWithKey` with the appropriate type seems to take care of the problem: this compiles fine
> 
> ```rescript
> module C = {
>   @react.component
>   let make = () => React.null
> }
> 
> module D = {
>   @react.component
>   let make = (~x="", ~y="") => React.string(x++y)
> }
> 
> let createElementWithKey = (comp, props, key) => React.createElement(comp, Jsx.addKeyProp(props, key))
> 
> let c1 = React.createElement(C.make, {})
> let c2 = createElementWithKey(C.make, {}, "")
> let d1 = React.createElement(D.make, {})
> let d2 = createElementWithKey(D.make,{}, "")
> let d3 = createElementWithKey(D.make, {x:""}, "")
> ```
> 
> So another way to go is to expose e.g. `Jsx. createElementWithKey` of the appropriate type instead of `Jsx.addKeyProp`. The type being `(React.component<'a>, 'a, string) => React.element`.

Wow! It looks more clean. I’ll fix it this way.

mattdamon108:
It is working nicely in the example project.
https://github.com/rescript-lang/syntax/pull/641

PR #643, opened 2022-09-19
add API for parsing impl/interface from a raw string

zth:
This adds an API for parsing impl/interface from a raw string. Main use case right now is in https://github.com/rescript-lang/rescript-vscode/pull/547 where I need to parse a raw string and don't want to have to put the contents into a temp file just to parse it.

Tried adding it to `parseEngine` first but that required me to add implementations for re/ml, which felt unecessary.


cristianoc (approved):

zth:
The failing Windows test seem to be about a cygwin mirror that times out. Merging anyway since that's expected to resolve itself soon.

PR #647, opened 2022-09-22
Unify comment formatting inside the empty blocks

ah-yu:
Unify the format of comments inside the empty block 
```res
// before
let foo = [// hi
]

let foo = [//hi
/* hi */]

// now -- same as prettier
let foo = [
  // hi
]

let foo = [
  // hi
 /* hi */
]

```


tests/printer/comments/expected/typexpr.res.txt
```diff
@@ -28,6 +28,10 @@ type t = /* c0 */ string /* c1 */ as 'x // after
 // Ptyp_poly
 type fn = {f: /* c0 */ 'a /* c1 */ 'b /* c2 */. /* c3 */ string /* c4 */}
 
+type fn = {/* comment1 */
```
IwanKaramazow, line 31: Can we tweak to output to look more like the input:
```
type fn = {
   /* comment1 */
   // comment2
 }
```

tests/printer/comments/expected/expr.res.txt
```diff
@@ -107,6 +107,11 @@ let user = /* before */ {
   /* c4 */ age /* c5 */: /* c6 */ 31 /* c7 */,
 } // after
 
+// Pexp_record with empty record
+let user = /* before */ {/* comment1 */
```
IwanKaramazow, line 111: The comments inside look a bit suboptimal, can we go for:
```
{
   /* comment1 */
   // comment2
 }
```

src/res_comments_table.ml
```diff
@@ -692,9 +692,12 @@ and walkTypeDeclaration (td : Parsetree.type_declaration) t comments =
     | Ptype_abstract | Ptype_open -> rest
     | Ptype_record labelDeclarations ->
       let () =
-        walkList
-          (labelDeclarations |> List.map (fun ld -> LabelDeclaration ld))
-          t rest
+        if List.length labelDeclarations = 0 then
```
cknitt, line 695: `List.length` iterates over the list unnecessarily, please use `match labelDeclarations with [] -> ...` instead.

cknitt:
Also, I just fixed CI, please rebase on latest master to get CI to run.

tests/printer/comments/expected/jsx.res.txt
```diff
@@ -14,9 +14,9 @@ module Cite = {
 
 <A /* comment */ />
 
-<A
-// Comment
-/>
+<A>
```
ah-yu, line 17: Could we keep tags opening if these are comments inside? @cknitt 
I found this change was introduced in https://github.com/rescript-lang/syntax/commit/e913c4cf9881efa8d91778e72cc50a61ce651b78

cknitt: Fine with me! That would make it easier to add code after the comment anyway.

cristianoc:
Has the scope of this PR changed?
There seem to be differences outside empty records.

ah-yu:
Yes. Now this PR is for adjusting the comments format inside an empty block(list, record, etc)

cristianoc:
> Yes. Now this PR is for adjusting the comments format inside an empty block(list, record, etc)

Would you add an entry in the CHANGELOG.md as this is user-visible.
@IwanKaramazow any more comments?

IwanKaramazow (commented):
If a user wrote `[/* a */]`, can we keep it on one line? Similar for the other "empty nodes".

We tend to look at the style of the author: did he write it over multiple lines or not?
* If it is on one line, keep it on one line (unless it breaks the column width).
* If it is written over multiple lines, force break it over multiple lines.

ah-yu:
That sounds reasonable, will fix it later.

src/res_printer.ml
```diff
@@ -705,19 +738,15 @@ and printModType ~customLayout modType cmtTbl =
...
-           [
-             Doc.lbrace;
-             Doc.indent
-               (Doc.concat
-                  [Doc.softLine; printCommentsInside cmtTbl modType.pmty_loc]);
-             Doc.softLine;
-             Doc.rbrace;
-           ])
+      let doc = printCommentsInside cmtTbl modType.pmty_loc in
+      if Doc.isNil doc then
```
cristianoc, line 742: This does the job but looks like tight coupling at a distance.
When is it that `printCommentsInside` currently produces `Nil`, and can we get hold of that information instead (e.g. return an optional).

cristianoc (approved):

src/res_printer.ml
IwanKaramazow: Yes, in an ideal world, we shouldn't inspect the contents of a doc to make formatting decisions. The information we go off is the AST (+ comments).

@ah-yu is the `isNil` absolutely necessary?

ah-yu: `isNil` is only used in the development phase, then I could do a quick test without too many code changes.

Forgot to clean it up and will do it later.

src/res_printer.ml
```diff
@@ -113,6 +113,12 @@ let hasNestedJsxOrMoreThanOneChild expr =
   in
   loop false expr
 
+let hasCommentsInside tbl loc =
+  match Hashtbl.find tbl.CommentTable.inside loc with
```
cristianoc, line 117: Can use find_opt

cristianoc:
@ah-yu just ping when this is ready.

ah-yu:
@cristianoc done

cristianoc:
> @cristianoc done

Thanks a lot for contributing this!

PR #648, opened 2022-09-22
Fix CI build

cknitt:


cristianoc (approved):
Thanks for the fix, been causing issues for a few days now!

PR #649, opened 2022-09-23
Update v4 doc about spread props

mattdamon108:
📝


cristianoc (approved):
Thanks.

cristianoc:
I'm thinking of fusing the 2 .md docs into one, which looks more like a guide.

Eg:
How to turn on V4.
Configuration and upgrade.
Details of the new transforms.

PR #650, opened 2022-09-23
Unify JSX docs in a single file.

cristianoc:


cristianoc:
@mattdamon108 would you review the updated, unified docs?

mattdamon108:
Maybe it is too much though, how about adding some kind of troubleshooting or migration how-to section. e.g. the handwritten component as per v3 before.

cristianoc:
> Maybe it is too much though, how about adding some kind of troubleshooting or migration how-to section. e.g. the handwritten component as per v3 before?

Would you list situations not currently covered but which should be covered, each with an example?

mattdamon108:
> > Maybe it is too much though, how about adding some kind of troubleshooting or migration how-to section. e.g. the handwritten component as per v3 before?
> 
> Would you list situations not currently covered but which should be covered, each with an example?

Sure, during test with community, there are some expected cases. I’ll list up soon.

PR #652, opened 2022-09-24
Update JSX v4 upgrade doc

mattdamon108:
This PR adds the upgrade doc for JSX v4. The upgrade doc lists the expected use cases which are incompatible with V3 and shows how to fix them.


cristianoc:
This file does not exist anymore, would you. integrated this inside JSXV4.md and adapt the structure as required?

mattdamon108:
BTW, it is intended. I thought the upgrade doc doesn't seem to fit in the spec doc.
But No problem, I'll add it to the spec doc.

mattdamon108:
Ah, I didn't noticed before https://github.com/rescript-lang/syntax/blob/9038cfbc2efec238eacb2034bf27b27df0d2f915/cli/JSXV4.md?plain=1#L16

cli/JSXV4.md
```diff
@@ -76,15 +76,95 @@ module Wrapper = {
...
+
+```rescript
+// V4
+module M = {
+  type props<'msg> = {msg: 'msg}
+  let make = props => <div> {React.string(props.msg)} </div>
+}
+```
+
+#### `React.Context`
```
cristianoc, line 109: @mattdamon108 would you clarify this example?
Is it necessary to depend on the internal representation in order to use `React.Context`?
If not, why is the fix not to avoid using the internal representation?

cli/JSXV4.md
```diff
@@ -76,15 +76,95 @@ module Wrapper = {
...
+      />
+      children
+    </div>
+  )
+}
+
+@react.component
+let make = () => {
+  <div>
+    <FancyInput ref=input> // only `ref` is allowed
```
cristianoc, line 152: @mattdamon108 looks like `input` here is a variable not defined anywhere?
Not sure I understand the example.

mattdamon108: I added https://github.com/rescript-lang/syntax/pull/652/commits/966b7e7059831fccad1a0553dcae5e94400719e1

cli/JSXV4.md
mattdamon108: I saw the error report from the community and the official doc shows the example of using internal representation to use `React.Context`. Therefore, many users might use this code and are expected to face the error with v3.

cristianoc: @mattdamon108 what's the answer to my question?
Should the solution in V4 use the V4 runtime representation, or should it not depend on the runtime representation?

mattdamon108: > @mattdamon108 what's the answer to my question? Should the solution in V4 use the V4 runtime representation, or should it not depend on the runtime representation?

The solution should use the v4 runtime representation.

cli/JSXV4.md
```diff
@@ -76,15 +76,95 @@ module Wrapper = {
...
+  let context = React.createContext(() => ())
+
+  module Provider = {
+    let provider = React.Context.provider(context)
+
+    @react.component
+    let make = (~value, ~children) => {
+      React.createElement(provider, {"value": value, "children": children}) // Error
+      // Fix : use the record
+      React.createElement(provider, {value, children}) // OK
```
mattdamon108, line 122: Expand the example seperately? Thought?

cli/JSXV4.md
```diff
@@ -145,19 +147,26 @@ module FancyInput = {
...
+  @react.component
+  let make = () => {
+    let input = React.useRef(Js.Nullable.null)
+
+    <div>
+      <FancyInput ref=input> // only `ref` is allowed
+        <button onClick> {React.string("Click to focus")} </button>
+      </FancyInput>
+    </div>
+  }
```
mattdamon108, line 165: I guess you miss-read my example. This module is not the fix of `FancyInput`

mattdamon108: The point of this example is that `ref` is only allowed, not `ref_` like the official doc. https://rescript-lang.org/docs/react/latest/forwarding-refs#discouraged-reactforwardref

cli/JSXV4.md
cristianoc: > > @mattdamon108 what's the answer to my question? Should the solution in V4 use the V4 runtime representation, or should it not depend on the runtime representation?
> 
> The solution should use the v4 runtime representation.

How about this instead:
```res
    module Prov = {
      let make = React.Context.provider(context)
    }
    ...
    <Prov value children /> // OK
```

Does it work?

mattdamon108: Let me test it.

cli/JSXV4.md
cristianoc: > The point of this example is that `ref` is only allowed, not `ref_` like the official doc. https://rescript-lang.org/docs/react/latest/forwarding-refs#discouraged-reactforwardref

OK would you fix the doc directly? Seems easier.

cli/JSXV4.md
mattdamon108: > Does it work?

Wow, it works. But, I'm bit unsure, because the js output different. I prefer and think the js output with internal representation is fine. But yes it works.

cristianoc: > > Does it work?
> 
> Wow, it works. But, I'm bit unsure, because the js output different. I prefer and think the js output with internal representation is fine. But yes it works without it.

Sounds good.
I will let you have one final pass to the examples, and then will merge.
OK?

cli/JSXV4.md
mattdamon108: https://github.com/rescript-lang/syntax/pull/652/commits/3e17c7bfef05d741c907efcabf53e0855b7526fa

cli/JSXV4.md
mattdamon108: https://github.com/rescript-lang/syntax/pull/652/commits/22ec394d56b800b3446af0f6a33334fa4625f6f8

cli/JSXV4.md
mattdamon108: Same goes here https://github.com/rescript-lang/syntax/pull/652/commits/22ec394d56b800b3446af0f6a33334fa4625f6f8

cristianoc:
@mattdamon108 ready to merge now?

mattdamon108:
Yes! 👍

PR #655, opened 2022-09-24
Better error msg and loc for multiple props type declaration

### LINKED ISSUE #654: Multiple definition of type name props when migriting to JSXv4
I had my own `type props = ...` (unrelated to React props, just using that name). Migrating to JSXv4 caused the following error:
```
rescript: [30/31] src/SchemaEditor-LumiStoreEditor.cmj
FAILED: src/SchemaEditor-LumiStoreEditor.cmj

  We've found a bug for you!
  SchemaEditor

  Multiple definition of the type name props.
  Names must be unique in a given structure or signature.

FAILED: cannot make progress due to previous errors.
```

The error seems valid (I should rename my own type), but:
* We may have to add this to the migration guide (JSXv4 claims the type name `props`)
* The error has no source location, which also makes clicking it in VSCode display an error

mattdamon108:
Adding loc to the generated props type as pointing to the same location of `@react.component` in each str_item and sig_item.

Fixed #654

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Fix location issue in error messages with JSX V4 where the body of the component is an application https://github.com/rescript-lang/syntax/pull/633
 - Fix issue where the printer would omit attributes for `->` and `|>` https://github.com/rescript-lang/syntax/pull/629
 - Fix printing of optional fields in records https://github.com/rescript-lang/rescript-compiler/issues/5654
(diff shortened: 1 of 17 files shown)


mattdamon108:
https://github.com/rescript-lang/syntax/pull/655/commits/3c7dc9defd762ff2e44f3d6dab1e59f9b49da2fa diff seems from the formatting change

cristianoc (commented):
Would you update the changelog?

mattdamon108:
Updated changelog https://github.com/rescript-lang/syntax/pull/655/commits/ff4bf0240bcfe2081aa79212bfe7d37dca550b13

PR #660, opened 2022-09-28
Fix location issue in make function in JSX V4 that breaks dead code e…

cristianoc:
…limination


cristianoc:
CC @IwanKaramazow

IwanKaramazow:
Thanks

Issue #663, opened 2022-09-28
Formatter erasing comments


nireno:
[...]

PR #664, opened 2022-09-29
fix: Avoid tail comments inside tag from being eaten

ah-yu:
Closes https://github.com/rescript-lang/syntax/issues/663

```res
// The minimal reproducible example
<A
a="1"
// comment
>
  <B />
</A>
```

tweak the output of `jsx` tag
```res
// before
<A
// comment  
  a="1"
/* comment */> 
  <B />
</A>

// now
<A
  // comment -- align with a="1"
  a="1"
  /* comment */   //  align with a="1"
>       // print > on a separate line -- same as prettier
  <B />
</A>

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@
 - Fix location issue in error messages with JSX V4 where the multiple props types are defined https://github.com/rescript-lang/syntax/pull/655
 - Fix location issue in make function in JSX V4 that breaks dead code elimination https://github.com/rescript-lang/syntax/pull/660
 - Fix parsing (hence pretty printing) of expressions with underscore `_` and comments.
+- Fix printing of comments inside JSX tag https://github.com/rescript-lang/syntax/pull/664
+- Fix issue where formatter erases tail comments inside JSX tag https://github.com/rescript-lang/syntax/issues/663
 
 ## ReScript 10.0
 
--- a/src/res_comments_table.ml
+++ b/src/res_comments_table.ml
@@ -1333,12 +1333,11 @@ and walkExpression expr t comments =
       | None -> ()
       | Some childrenExpr ->
         let leading, inside, _ = partitionByLoc after childrenExpr.pexp_loc in
-        if props = [] then (
+        if props = [] then
           let afterExpr, _ =
             partitionAdjacentTrailing callExpr.pexp_loc after
           in
-          attach t.trailing callExpr.pexp_loc afterExpr;
-          walkExpression childrenExpr t inside)
+          attach t.trailing callExpr.pexp_loc afterExpr
         else
           walkList (props |> List.map (fun (_, e) -> ExprArgument e)) t leading;
         walkExpression childrenExpr t inside)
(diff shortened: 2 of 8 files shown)


src/res_comments_table.ml
```diff
@@ -1310,9 +1311,36 @@ and walkExpression expr t comments =
...
+               | Asttypes.Nolabel -> false
+               | _ -> true)
+      in
+      let maybeChildren =
+        arguments
+        |> List.find_opt (fun (label, _) ->
+               label = Asttypes.Labelled "children")
+      in
+      match maybeChildren with
+      | None -> ()
```
cristianoc, line 1329: Is this not walking the props at all?
What happens if you have props with comments and no children?

src/res_comments_table.ml
```diff
@@ -1310,9 +1311,36 @@ and walkExpression expr t comments =
...
+      | Some (_, children) ->
+        let leading, inside, _ = partitionByLoc after children.pexp_loc in
+        if props = [] then
+          let afterExpr, _ =
```
cristianoc, line 1333: What happen to the second argument, `_` in this case?
Why can it be ignored?

CHANGELOG.md
```diff
@@ -31,6 +31,8 @@
 - Fix location issue in error messages with JSX V4 where the multiple props types are defined https://github.com/rescript-lang/syntax/pull/655
 - Fix location issue in make function in JSX V4 that breaks dead code elimination https://github.com/rescript-lang/syntax/pull/660
 - Fix parsing (hence pretty printing) of expressions with underscore `_` and comments.
+- Fix printing of comments inside JSX tag https://github.com/rescript-lang/syntax/pull/664
```
cristianoc, line 34: Oh this must have been missed somewhere.

cristianoc (commented):
Some questions in the code.

src/res_comments_table.ml
ah-yu: Would you give me a hint under what circumstances this would happen? I tested some cases and found children were always there

If there are no children elements inside, there will be `Pexp_construct "[]" ` in the AST tree
 ```
Labelled "children"
          expression (test.res[2,1+2]..[2,1+3])
            Pexp_construct "[]" (test.res[2,1+2]..[2,1+3]) ghost
            None
```

src/res_comments_table.ml
ah-yu: Trailing comments is always an empty list, so there is no need to deal with it.

CHANGELOG.md
ah-yu: This change is also included in this PR. See details in the PR description

cristianoc: OK thanks.

src/res_comments_table.ml
cristianoc: How about adding a comment saying just that.

src/res_comments_table.ml
cristianoc: How about adding a comment saying that.

cristianoc (approved):
Looks great.
Just asking to add a couple of comments to it's clear the code paths not covered are intentional and not an oversight.

ah-yu:
@cristianoc Comments added.

PR #666, opened 2022-10-01
Fix JSX v4 for first class module

mattdamon108:
Feedback from forum https://forum.rescript-lang.org/t/call-for-help-2-test-jsx-v4/3781/16?u=moondaddi

This PR fixes it


cristianoc (approved):
Looking great.
Would you update the changelog?

mattdamon108:
Does it go into master or 10.1_release?

PR #668, opened 2022-10-02
Remove printing of arrow functions with "multiple" dots

IwanKaramazow:
Arrow functions with multiple dots `(. a, . b) => a + b` are considered an anti-pattern. This syntax *will be removed* in the future. As a first step, the printer will know print this as multiple arrow functions. (Like in JavaScript)

**Before**
```rescript
let f = (. a, . b) => a + b
type f = (. int, . int) => int
```

**After**
```rescript
let f = (. a) => (. b) => a + b
type f = (. int) => (. int) => int
```

### DIFF
--- a/src/res_parsetree_viewer.ml
+++ b/src/res_parsetree_viewer.ml
@@ -11,7 +11,7 @@ let arrowType ct =
       process attrsBefore (arg :: acc) typ2
     | {
      ptyp_desc = Ptyp_arrow (Nolabel, _typ1, _typ2);
-     ptyp_attributes = [({txt = "bs" | "res.async"}, _)];
+     ptyp_attributes = [({txt = "bs"}, _)];
     } ->
       (* stop here, the uncurried attribute always indicates the beginning of an arrow function
          * e.g. `(. int) => (. int)` instead of `(. int, . int)` *)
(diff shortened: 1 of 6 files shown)


src/res_parsetree_viewer.ml
```diff
@@ -10,11 +10,12 @@ let arrowType ct =
       let arg = ([], lbl, typ1) in
       process attrsBefore (arg :: acc) typ2
     | {
-     ptyp_desc = Ptyp_arrow ((Nolabel as lbl), typ1, typ2);
-     ptyp_attributes = [({txt = "bs" | "res.async"}, _)] as attrs;
+     ptyp_desc = Ptyp_arrow (Nolabel, _typ1, _typ2);
+     ptyp_attributes = [({txt = "bs" | "res.async"}, _)];
```
cristianoc, line 14: Can you add examples of nested async functions?

IwanKaramazow: Not sure what that is doing here. `type f = async (…) => …` does not exist. Will remove.

IwanKaramazow: Done in https://github.com/rescript-lang/syntax/pull/668/commits/def9fa087ac93f0b9857e0e20c2de4fd6b96ce82

cristianoc (approved):

cristianoc:
Would you add a line to the changelog and merge.

PR #676, opened 2022-10-13
Fix eaten async for function with labelled args

### LINKED ISSUE #674: async gets formatted away for functions with labeled arguments
```rescript
let f = async (a, b) => a + b

let f = async (~a, ~b) => a + b
```
gets reformatted to 
```rescript
let f = async (a, b) => a + b

let f = (~a, ~b) => a + b
```

mattdamon108:
This PR fixes #674

### DIFF
--- a/src/res_parsetree_viewer.ml
+++ b/src/res_parsetree_viewer.ml
@@ -73,6 +73,13 @@ let processFunctionAttributes attrs =
   in
   process false false [] attrs
 
+let hasAsyncAttribute attrs =
+  List.exists
+    (function
+      | {Location.txt = "res.async"}, _ -> true
+      | _ -> false)
+    attrs
+
 let hasAwaitAttribute attrs =
   List.exists
     (function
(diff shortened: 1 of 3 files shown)


mattdamon108:
I don't know the history of why letting the pexp_fun with nolabel arg keeps the attributes. https://github.com/rescript-lang/syntax/blob/a3c27ef3011e1c37ee3c2fbf8229f4dca49d3de7/src/res_parsetree_viewer.ml#L176

mattdamon108:
This PR allows the Pexp_fun with labeled and optional to keep the `res.async` attribute only.

cknitt (approved):
Looks good to me, thanks a lot for fixing this! 🎉

(I don't know why Pexp_fun with Nolabel keeps all attributes either though.)

cristianoc:
Oops, I had another fix done on the plane:
https://github.com/rescript-lang/syntax/pull/677

cristianoc:
Basically, for `(@foo ~a) => ...` when one wants to put an attribute on the label, there's no way to represent this directly in the AST, as labelled args don't have attributes.
So the convention is to take the attributes on the function, and consider them applied to the label.

This does not fit well with the fact that `async` is put on the function, and some adjustment is needed.

src/res_parsetree_viewer.ml
```diff
@@ -177,6 +184,17 @@ let funExpr expr =
...
+      pexp_desc = Pexp_fun (_, _defaultExpr, _pattern, _returnExpr);
+      pexp_attributes = attrs;
+    } as expr
+    when hasAsyncAttribute attrs ->
+    let asyncAttr =
+      List.filter
+        (fun (({txt}, _) : Parsetree.attribute) -> txt = "res.async")
+        attrs
+    in
+    collect asyncAttr [] {expr with pexp_attributes = []}
```
cristianoc, line 197: This throws away all the other attributes, if there are some.

mattdamon108:
> 

There was a reason! Thank you for the explanation. I couldn't guess it. 👍

src/res_parsetree_viewer.ml
mattdamon108: You are correct.

mattdamon108:
I'm going to close this PR, #677 will fix the issue.

PR #677, opened 2022-10-13
Fix issue where the formatter would delete `async` in a function with…

cristianoc:
… labelled arguments.

### DIFF
--- a/src/res_parsetree_viewer.ml
+++ b/src/res_parsetree_viewer.ml
@@ -167,8 +167,15 @@ let funExpr expr =
          (((Labelled _ | Optional _) as lbl), defaultExpr, pattern, returnExpr);
      pexp_attributes = attrs;
     } ->
-      let parameter = Parameter {attrs; lbl; defaultExpr; pat = pattern} in
-      collect attrsBefore (parameter :: acc) returnExpr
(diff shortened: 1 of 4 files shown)


src/res_parsetree_viewer.ml
```diff
@@ -167,8 +167,15 @@ let funExpr expr =
          (((Labelled _ | Optional _) as lbl), defaultExpr, pattern, returnExpr);
      pexp_attributes = attrs;
     } ->
-      let parameter = Parameter {attrs; lbl; defaultExpr; pat = pattern} in
-      collect attrsBefore (parameter :: acc) returnExpr
+      (* Normally attributes are attached to the labelled argument, e.g. (@foo ~x) => ...
+         In the case of `@res.async`, pass the attribute to the outside *)
```
mattdamon108, line 171: Ah, there was a reason! I'm going to close my #676

mattdamon108: How about adding tests for this case? Maybe,
```rescript
let f = async (@attr1 ~x, @attr2 ~y) => x + y
```

cristianoc: Good point.

Issue #679, opened 2022-10-14
Parentheses added for async function with single param


cknitt:
```rescript
let f = async a => a
```
gets formatted to

```rescript
let f = async (a) => a
```

Is this intentional? Because for non-async functions we are *not* adding parentheses automatically, but keep `let f = a => a`.

mattdamon108:
Oops, I'm sorry https://github.com/rescript-lang/syntax/issues/674 had two different issues there.

cristianoc:
Looking.

cristianoc:
There are in fact several async examples requiring attention https://github.com/rescript-lang/syntax/pull/680

PR #680, opened 2022-10-14
Fix several printing issues with `async` including an infinite loop.

cristianoc:
See https://github.com/rescript-lang/syntax/issues/679

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 - Fix issue where the JSX prop has type annotation of the first class module https://github.com/rescript-lang/syntax/pull/666
 - Fix issue where a spread `...x` in non-last position would not be reported as syntax error https://github.com/rescript-lang/syntax/pull/673/
 - Fix issue where the formatter would delete `async` in a function with labelled arguments.
+- Fix several printing issues with `async` including an infinite loop https://github.com/rescript-lang/syntax/pull/680
 
 #### :eyeglasses: Spec Compliance
 
--- a/src/res_parsetree_viewer.ml
+++ b/src/res_parsetree_viewer.ml
@@ -136,7 +136,7 @@ let funExpr expr =
       collectNewTypes (stringLoc :: acc) returnExpr
     | returnExpr -> (List.rev acc, returnExpr)
   in
-  let rec collect attrsBefore acc expr =
+  let rec collect n attrsBefore acc expr =
     match expr with
     | {
      pexp_desc =
@@ -152,13 +152,17 @@ let funExpr expr =
      pexp_attributes = [];
     } ->
       let parameter = Parameter {attrs = []; lbl; defaultExpr; pat = pattern} in
-      collect attrsBefore (parameter :: acc) returnExpr
+      collect (n + 1) attrsBefore (parameter :: acc) returnExpr
     | {pexp_desc = Pexp_newtype (stringLoc, rest); pexp_attributes = attrs} ->
       let stringLocs, returnExpr = collectNewTypes [stringLoc] rest in
       let param = NewTypes {attrs; locs = stringLocs} in
-      collect attrsBefore (param :: acc) returnExpr
-    | {pexp_desc = Pexp_fun _; pexp_attributes = [({txt = "bs"}, _)]} ->
-      (* stop here, the uncurried attribute always indicates the beginning of an arrow function
+      collect (n + 1) attrsBefore (param :: acc) returnExpr
+    | {pexp_desc = Pexp_fun _; pexp_attributes}
+      when pexp_attributes
+           |> List.exists (fun ({Location.txt}, _) ->
+                  txt = "bs" || txt = "res.async")
+           && n > 0 ->
+      (* stop here, the uncurried or async attribute always indicates the beginning of an arrow function
        * e.g. `(. a) => (. b)` instead of `(. a, . b)` *)
       (attrsBefore, List.rev acc, expr)
     | {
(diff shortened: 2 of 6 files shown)


mattdamon108:
The pattern matching seems complicated and presumably the Pexp_fun with labeled arg can have attributes, not with Nolabel arg. Is it syntax that only Pexp-fun with Labeled and Optional arg can have attributes?

cristianoc:
> 

Just trying to respect what seem to be the existing conventions. And yes it looks complicated. The attributes I think can be used in JSX to specify `@as` for arguments right?

mattdamon108:
> Just trying to respect what seem to be the existing conventions. And yes it looks complicated. The attributes I think can be used in JSX to specify `@as` for arguments right?

In user space? Not that I know. The attribute for arg is mainly used in `Pstr_primitives`, I mean the external. We use the attributes such as `@as` often when we bind to components or functions. The attribute for `Pexp_fun` is not used as long as I know.
But I see that the output of JSX V3 has `@warning` attributes for arg.

mattdamon108:
In order to make the pattern matching more simple, then I look through several cases of attributes with fun args. I found something strange between labeled and nolabel arg.

```rescript
let f7 = (@attr ~a, @attr ~b) => ()
let f8 = (@attr a, @attr b) => ()
```
```
// (f7) @attr ~a
// @attr is parsed to pexp_attributes
        expression (tests/printer/expr/fun2.res[9,244+9]..[9,244+35])
          attribute "attr" (tests/printer/expr/fun2.res[9,244+10]..[9,244+15])
            []
          Pexp_fun
          Labelled "a"
          None
          pattern (tests/printer/expr/fun2.res[9,244+10]..[9,244+18])
            attribute "ns.namedArgLoc" (tests/printer/expr/fun2.res[9,244+17]..[9,244+18])
              []
            Ppat_var "a" (tests/printer/expr/fun2.res[9,244+10]..[9,244+18])

// (f8) @attr a
// @attr is pared to ppat_attributes
        expression (tests/printer/expr/fun2.res[10,280+9]..[10,280+33])
          Pexp_fun
          Nolabel
          None
          pattern (tests/printer/expr/fun2.res[10,280+16]..[10,280+17])
            attribute "attr" (tests/printer/expr/fun2.res[10,280+10]..[10,280+15])
              []
            Ppat_var "a" (tests/printer/expr/fun2.res[10,280+16]..[10,280+17])
```
Can we parse the attributes for arg to the `ppat_attributes` of arg for both cases? The only `bs` and `res.async` attributes are parsed to `pexp_attributes` in `Pexp_fun`. Then, we may be able to make the more simple implementation of the printer.

cristianoc:
The one thing I'm really not sure about is why labelled and unlabelled args are treated differently.
Wondering whether there's something that might make use of the difference.

Also, this
```res
let f12 = @a (@b x) => 3
```
formats to
```
let f12 = @a x => 3
```

mattdamon108:
I think we need to check this decision history.
https://github.com/rescript-lang/syntax/blob/master/src/res_core.ml#L1537:L1542

cristianoc:
> In order to make the pattern matching more simple, then I look through several cases of attributes with fun args. I found something strange between labeled and nolabel arg.
> 
> ```rescript
> let f7 = (@attr ~a, @attr ~b) => ()
> let f8 = (@attr a, @attr b) => ()
> ```
> 
>  ```
> // (f7) @attr ~a
> // @attr is parsed to pexp_attributes
>         expression (tests/printer/expr/fun2.res[9,244+9]..[9,244+35])
>           attribute "attr" (tests/printer/expr/fun2.res[9,244+10]..[9,244+15])
>             []
>           Pexp_fun
>           Labelled "a"
>           None
>           pattern (tests/printer/expr/fun2.res[9,244+10]..[9,244+18])
>             attribute "ns.namedArgLoc" (tests/printer/expr/fun2.res[9,244+17]..[9,244+18])
>               []
>             Ppat_var "a" (tests/printer/expr/fun2.res[9,244+10]..[9,244+18])
> 
> // (f8) @attr a
> // @attr is pared to ppat_attributes
>         expression (tests/printer/expr/fun2.res[10,280+9]..[10,280+33])
>           Pexp_fun
>           Nolabel
>           None
>           pattern (tests/printer/expr/fun2.res[10,280+16]..[10,280+17])
>             attribute "attr" (tests/printer/expr/fun2.res[10,280+10]..[10,280+15])
>               []
>             Ppat_var "a" (tests/printer/expr/fun2.res[10,280+16]..[10,280+17])
> ```
> 
> Can we parse the attributes for arg to the `ppat_attributes` of arg for both cases? The only `bs` and `res.async` attributes are parsed to `pexp_attributes` in `Pexp_fun`. Then, we may be able to make the more simple implementation of the printer.

Yes this makes sense, and should respect the intent of the original code. Basically attributes outside `( )` would be added to the function, and inside `(...)` to the pattern.
And `async` goes to the function, and uncurried `(. ...)` also go to the function.

Is this ^ what you have in mind?

mattdamon108:
> Yes this makes sense, and should respect the intent of the original code. Basically attributes outside `( )` would be added to the function, and inside `(...)` to the pattern. And `async` goes to the function, and uncurried `(. ...)` also go to the function.
> 
> Is this ^ what you have in mind?

Exactly!
But I'm a little worrying about the reason for the decision to treat differently between labeled arg and nolabed arg. Any side-effect after changing it

cristianoc:
I suspect the distinction was not really intentional, and things just developed this way.
@IwanKaramazow do you remember?

IwanKaramazow:
The distinction was indeed not really intentional, I guess no ReScript developer ever hit this path.

> Basically attributes outside ( ) would be added to the function, and inside (...) to the pattern.
And async goes to the function, and uncurried (. ...) also go to the function.

Exactly! This is also my mental model.

cristianoc:
Proposal, I'll add a few more examples and merge this PR. Will go into 10.1 as it needs to fix async.

Then create an issue about the proper solution discussed here, to go into v11, so there's enough time to figure out if it affects anything.

cristianoc:
https://github.com/rescript-lang/rescript-compiler/issues/5735

Issue #675, opened 2022-10-13
Unnecessary format changes in JSX between 10.0.1 and 10.1.0-rc.1


cknitt:
I just installed 10.1.0-rc.1 in a project of ours for testing and ran the formatter across the code. This resulted in **83 files changed**.

The reason was that as a side effect of #664, JSX tags are now formatted differently, as can also be seen in the test result diffs in this PR.

Before:

```rescript
<svg
  viewBox="0 0 24 24"
  width="24"
  height="24"
  stroke="currentColor"
  strokeWidth="2"
  fill="none"
  strokeLinecap="round"
  strokeLinejoin="round"
  className="text-gray-400">
  <polyline points="6 9 12 15 18 9" />
</svg>
```

After:

```rescript
<svg
  viewBox="0 0 24 24"
  width="24"
  height="24"
  stroke="currentColor"
  strokeWidth="2"
  fill="none"
  strokeLinecap="round"
  strokeLinejoin="round"
  className="text-gray-400"
>
  <polyline points="6 9 12 15 18 9" />
</svg>
```

While it may be debatable which style is better, I think the difference is not worth causing so many format changes in users' codebases.

ah-yu:
Hello, I investigated some popular formatter(prettier, deno format), and they put the `>` on a new line. Probably this code style is more familiar to Js users.

It seems like a breaking change maybe it is better released in the next major if you are in favour of this change.

IwanKaramazow:
> I investigated some popular formatter(prettier, deno format), and they put the > on a new line. Probably this code style is more familiar to Js users.

Don't forget our existing users. We should try to reserve this for v11 and restore the previous way of formatting.

cristianoc:
We need to revert this change on branch 10.1 and add it to master with a changelog listing as breaking change.

cristianoc:
Since this is the syntax repo, seems best to revert here and leave an open PR to add it back.
And an issue on v11 milestone to add it later.

cknitt:
Actually for Prettier this is a configuration option (`bracketSameLine`). It's true that `bracketSameLine = false` is Prettier's default though (whereas what we have in ReScript 10.0 corresponds to `bracketSameLine = true`).

However, we are not following Prettier's defaults in other places either, e.g. Prettier 2.0+ has `arrowParens = "always"` as a default whereas ReScript's formatting corresponds to `arrowParens = "avoid"`.

[...]

PR #685, opened 2022-10-16
fix: put > on a new line if the tag or last prop has trailing comments

ah-yu:
Closes https://github.com/rescript-lang/syntax/issues/675 again...

See https://github.com/rescript-lang/syntax/issues/675#issuecomment-1279882340


ah-yu:
@cknitt Could you try again on your codebase and see if that formatting issue is still there

cknitt:
Yes! The problem is now resolved. Thanks a lot for your work! 👍

Issue #687, opened 2022-10-16
await lost on reformat


cknitt:
From https://github.com/rescript-lang/syntax/issues/686#issuecomment-1279897564:

```rescript
(await (server->start))->ignore
```

loses the `await` on reformat:

```rescript
server->start->ignore
```

PR #688, opened 2022-10-16
Fix issue where printing nested pipe discards await

cristianoc:
Fixes https://github.com/rescript-lang/syntax/issues/687

Certain nested patterns for binary operators are printed "manually" so the special printing for await is not used.


cknitt (approved):
🎉

Issue #670, opened 2022-10-11
Feature: relax the constraint of `...` in list construction


bobzhang:
Currently, `...` is only allowed in the tail position, `list {1,2,... tail}`,
For example:
```res
let l = a => list{1,2,...a} // works
```
```res
let l = a => list{1,2,...a, 3,  ...a} // fails
```
The current situation is a bit weird, it is not a syntax error, but a type error 

`This has type: int
  Somewhere wanted: list<int>`
  
  This also seems to be a bug:
```
let l = (a,b) => list{1,2, ...b , 3,...a}
```
The inferred type for `b` is `int`

butterunderflow:
This issue seems raised from here. 
https://github.com/rescript-lang/syntax/blob/c2b5750f5fa8b9a50875ec7d6ff96319171f46a7/src/res_core.ml#L3727

The current behavior is: if the last list element expression is a [spread](https://rescript-lang.org/docs/manual/latest/array-and-list#immutable-prepend), the parser will not check the rest element whether contains a spread expression.  So the list expression 
```reason
list{...a, ...b ,...a}
```
would be parsed to OCaml
```reason
a :: b :: a
```
which raises a type error in subsequent type checking. 

But 
```reason
list{...a, ...b, c}
```
would raise a syntax error, because the last is not a spread.

My [commit]() seems fix this.

Would someone help review my code?

cristianoc:
Would you create a pull request?

bobzhang:
@butterunderflow 
A proposal for fix:
when `...` only appears in the tail position, nothing changes, so no regressions would happen.

when multiple `...` appears

```
 list{ a0 , ... b0 , ai, ... bi, }
```
==> `Belt.List.concatMany ([ list{a0, ...b0},  list{ai,... bi}]`

This would make a working version first, we could improve the performance later

butterunderflow:
As far as I understand, after this feature is added, there will be no syntax errors related to wrong spread location, since spread expression at anywhere could be compiled to list concatenation. Do I understand correctly?

bobzhang:
yes

PR #692, opened 2022-10-20
Spread anywhere feature

butterunderflow:
Feature request from #670. 
The new feature will allow spread anywhere in a list creation expression.
For example, list creation like 
```rescript
let _ = list{1, 2, 3, ...x, 4, 5, ...y, 6}
```
will no longer trigger an error anymore, but parsed to
``` OCaml
let _ = Belt.List.concatMany [|(1 :: 2 :: 3 :: x);(4 :: 5 :: y);[6]|]
```

Can anyone help review this feature PR? Thanks ^v^


cristianoc:
Would you add printing tests?

butterunderflow:
What is `printing tests` stand for? Like printing some parse results related to this feature?

cristianoc:
Add a file in tests/print or extend an existing one where list examples are.

It will tell you how things print, and whether printing is idempotent, so you don't get different results if you format twice.

butterunderflow:
Thanks, I think I understand. I'll give it a try.

cristianoc:
See how the syntax of normal lists is preserved by printing.
A similar mechanism is needed for this extension to lists.

namenu:
tiny concern, but would it be better not allowing this expression at all ?

lists are not okay when manipulation is occurred at non-head position.

Once spreadings are used inadvertently, it would cause performance problem.

By now, list spreading guarantees constant time

butterunderflow:
IMHO, it's still a useful feature. For some reason.
First, IMHO again, list concatenation is a little bit verbose at present. We have to write some prefixes of Module.
```rescript
Belt.List.concatMany([list{1, 2, ...x}, list{3, ...x}, list{4, ...x}, list{5}])
// or 
Belt.List.concat(list{1, 2, ...x}, list{3, ...x})
// or 
List.concat(list{list{1, 2, ...x}, list{3, ...x}})
```
Second, I think if we support @ operator (like OCaml), the verbose issues will be moderated. But for the existence of list open token `list{`,  looks still not concise enough. Like we have to tell the reader repeatedly "it's a **list**".
```rescript
// pretend we have operator @ 
list{1, 2, ...x} @ list{3, ...x} @ list{4, ...x} @ list{5}
list{1,2, ...x} @ list{3, ...x}
```
If we let spread show anywhere, we can just write:
```rescript
list{1, 2, ...x, 3, ...x, 4, ...x, 5}
list{1, 2, ...x, 3, ...x}
```
I think that's more concise, right?

And, as far as I know, literally, the word "spread" itself is not intuitively associated with constant time operation, unlike "prepend".

Finally, the performance issue you mentioned does exist while this feature is being misused. But this feature will not downside any performance of existing code, and I think (**I'm not sure**) concatenate seems can be optimized in some way. So I think it's still a good deal.  I'll always wait for more suggestions.

IwanKaramazow:
Before adding this to the language, we should take a closer look at the design of this feature.
What is the reach of this problem? There are 3 vectors:
- volume of GH issues, forum posts, twitter etc…
- how blocking is this problem if the developer encounters
- how many of the total ReScript devs does it affect in their day to day work?

When making design decisions regarding new syntax in the language, it’s also important to weigh the costs, not all of which are performance, against the potential benefits. The following things can happen:
- The language gets larger.
- The language gets slower.
- The documentation becomes more extensive.
- We're spending time developing new syntax, rather than refining other parts of the language.
- We run the risk of introducing changes that don't play well with existing syntax. (Why don't we allow this for array then?)

bobzhang:
Let me explain a bit why this feature it is useful and why you should not worry about performance.

This feature is very helpful for expressing lots of  algorithms, for example 

```re
Add(e1,e2) => list {... compile(e1), ... compile(e2), Add }
```
It is much more readable than the alternatives, and as efficient as you could, in theory the compiler can do 
a better job than you since it kows the `compile` is a temporay variable and can do some smart mutations.

So the argumets are:
- You should avoid list if you care about performance
- If you want to concate two lists, the alternative is not better at performance, you have to express this pattern, by
disabling this syntax does not mean you can avoid the problem, JavaScript is also consistent here

Before we add this feature, this used to be a bug, the syntax was accepted but not reported error, it is worse so 
that we are not enriching the syntax: we are **fixing** existing issues.

With regard to developer resources, this is the feature I would like to add on day one when making BuckleScript, 
finally, I have some dedicated resources to make it possible, it also makes the sample code for [PL course in ReScript](https://forum.rescript-lang.org/t/pl-course-in-rescript/3818) much more readable, instructions concatenation are everywhere. 

Hope my explanantions make the motivation clear

tests/parsing/errors/other/expected/spread.res.txt
```diff
@@ -114,7 +100,7 @@ let arr = [|x;y|]
 let [|arr;_|] = [|1;2;3|]
 let record = { x with y }
 let { x; y } = myRecord
-let myList = x :: y
+let myList = Belt.List.concatMany [|x;y|]
```
bobzhang, line 103: - this case should be specialized?
- this case too `{ ... x}`

tests/printer/expr/expected/list.res.txt
```diff
@@ -2,6 +2,7 @@ let x = list{}
 let x = list{1}
 let x = list{1, 2}
 let x = list{1, 2, 3}
+let x = Belt.List.concatMany([list{1, 2, ...x}, list{3, ...x}])
```
bobzhang, line 6: could this be improved? print it as `list{ 1, 2, ...x , 3 , ...x}`

butterunderflow: It could be, but just by changing the printer can cause some problems. For example, list{. . x, . .y} and Belt.List.concatMany([x, y]) will be printed as the same thing. I am still trying to solve this problem.

bobzhang: can you add an attribute to Belt.List.concatMany to avoid accidental simplifications?

cristianoc: This is what is done for string templates, I believe, to differentiate from normal string concatenation.

cristianoc: Check what `isTemplateLiteral` does.

butterunderflow: Thank you for all your suggestions, I'll learn more about that.

IwanKaramazow:
Thanks for the context @bobzhang 👍

src/res_parsetree_viewer.ml
```diff
@@ -618,6 +618,26 @@ let isTemplateLiteral expr =
...
+let isSpreadBeltListConcat expr =
+  match expr.pexp_desc with
+  | Pexp_ident
+      {
+        txt =
+          Longident.Ldot
+            (Longident.Ldot (Longident.Lident "Belt", "List"), "concatMany");
+      }
+    when hasSpreadAttr expr.pexp_attributes ->
+    true
```
bobzhang, line 638: simplified: ` -> hasSpreadAttr exp.pexp_attributes`

tests/parsing/errors/other/expected/spread.res.txt
```diff
@@ -42,79 +42,64 @@ Explanation: since records have a known, fixed shape, a spread like `{a, ...b}`
...
-   9 │ 
-  10 │ type t = {...a}
+  5 │ let {...x, ...y} = myRecord
+  6 │ 
+  7 │ let list{...x, ...y} = myList
+  8 │ 
+  9 │ type t = {...a}
 
   List pattern matches only supports one `...` spread, at the end.
 Explanation: a list spread at the tail is efficient, but a spread in the middle would create new lists; out of performance concern, our pattern matching currently guarantees to never create new intermediate data.
```
bobzhang, line 62: we may need update the docs about pattern in later commits, it is not supported since it is not deterministic

tests/printer/expr/expected/list.res.txt
```diff
@@ -2,6 +2,8 @@ let x = list{}
 let x = list{1}
 let x = list{1, 2}
 let x = list{1, 2, 3}
+let x = list{1, 2, ...x, 3, ...x}
+let x = Belt.List.concatMany([list{1, 2, ...x}, [list{3, ...x}]])
 
 let x = list{
```
bobzhang, line 8: can you add some more test cases, some edge cases like:
```re
let x = list{ ...xs}
let x = list{1, ...xs}
let x = list{xs, ... ys}
let x = list{ ...xs, ... ys}
```

bobzhang:
@butterunderflow looks good to me once the comments are addressed.
Note `Belt.List.concatMany` is an implementation detail, in theory, we could swap more efficient ones in the future

butterunderflow:
Done. Please check again, thanks!

bobzhang:
@butterunderflow Thanks, cannot wait for to use it in the course!

cristianoc:
@butterunderflow would you add a line to CHANGELOG.md

butterunderflow:
Done. Please check https://github.com/rescript-lang/syntax/pull/698

PR #693, opened 2022-10-21
Fix react comp key type

mattdamon108:
This PR fixes https://github.com/rescript-lang/rescript-react/issues/73

### DIFF
--- a/cli/JSXV4.md
+++ b/cli/JSXV4.md
@@ -11,7 +11,7 @@ To build an entire project in V4 mode, including all its dependencies, use the n
 "jsx": { "version": 4 }
 ```
 
-> Note that JSX V4 requires the rescript compiler 10.1 or higher, and `rescript-react` version `0.11` or higher. In addition, `react` version `18.2` is required.
+> Note that JSX V4 requires the rescript compiler 10.1 or higher, and `rescript-react` version `0.11` or higher. In addition, `react` version `18.0` is required.
 
 ## Configuration And Upgrade
 
@@ -286,7 +286,11 @@ React.createElement(Comp.make, {x, y: 7, ?z})
 
 <Comp x key="7">
 // is transformed to
-React.createElement(Comp.make, Jsx.addKeyProp({x: x}, "7"))
+React.createElement(Comp.make, Jsx.addKeyProp(~key="7", {x: x}))
+
+<Comp x key=?Some("7")>
+// is transformed to
+React.createElement(Comp.make, Jsx.addKeyProp(~key=?Some("7"), {x: x}))
 ```
 
 ### New experimental automatic mode
--- a/cli/JSXV4.md
+++ b/cli/JSXV4.md
@@ -286,11 +286,11 @@ React.createElement(Comp.make, {x, y: 7, ?z})
 
 <Comp x key="7">
 // is transformed to
-React.createElement(Comp.make, Jsx.addKeyProp(~key="7", {x: x}))
+React.createElement(Comp.make, React.addKeyProp(~key="7", {x: x}))
 
 <Comp x key=?Some("7")>
 // is transformed to
-React.createElement(Comp.make, Jsx.addKeyProp(~key=?Some("7"), {x: x}))
+React.createElement(Comp.make, React.addKeyProp(~key=?Some("7"), {x: x}))
 ```
 
 ### New experimental automatic mode
--- a/tests/ppx/react/expected/externalWithCustomName.res.txt
+++ b/tests/ppx/react/expected/externalWithCustomName.res.txt
@@ -36,4 +36,4 @@ module Foo = {
   external component: React.componentLike<props<int, string>, React.element> = "component"
 }
 
-let t = React.jsxWithKey(Foo.component, {a: 1, b: "1"})
+let t = React.jsx(Foo.component, {a: 1, b: "1"})
--- a/tests/ppx/react/expected/externalWithCustomName.res.txt
+++ b/tests/ppx/react/expected/externalWithCustomName.res.txt
@@ -36,4 +36,4 @@ module Foo = {
   external component: React.componentLike<props<int, string>, React.element> = "component"
 }
 
-let t = React.jsx(Foo.component, {a: 1, b: "1"})
+let t = React.jsxWithKey(Foo.component, {a: 1, b: "1"})
--- a/tests/ppx/react/expected/forwardRef.res.txt
+++ b/tests/ppx/react/expected/forwardRef.res.txt
@@ -167,7 +167,10 @@ module V4A = {
       "div",
       {
         children: ?ReactDOM.someElement(
-          React.jsx(FancyInput.make, {ref: input, children: {React.string("Click to focus")}}),
+          React.jsxWithKey(
+            FancyInput.make,
+            {ref: input, children: {React.string("Click to focus")}},
+          ),
         ),
       },
     )
--- a/tests/ppx/react/expected/forwardRef.res.txt
+++ b/tests/ppx/react/expected/forwardRef.res.txt
@@ -167,10 +167,7 @@ module V4A = {
       "div",
       {
         children: ?ReactDOM.someElement(
-          React.jsxWithKey(
-            FancyInput.make,
-            {ref: input, children: {React.string("Click to focus")}},
-          ),
+          React.jsx(FancyInput.make, {ref: input, children: {React.string("Click to focus")}}),
         ),
       },
     )
--- a/tests/ppx/react/expected/spreadProps.res.txt
+++ b/tests/ppx/react/expected/spreadProps.res.txt
@@ -19,7 +19,7 @@ let c2 = React.createElement(A.make, {...p, x: "x"})
 // let c0 = <A x="x" {...p0} {...p1} />
 
 // only spread props
-let c1 = React.jsxWithKey(A.make, p)
+let c1 = React.jsx(A.make, p)
 
 // reversed order
-let c2 = React.jsxWithKey(A.make, {...p, x: "x"})
+let c2 = React.jsx(A.make, {...p, x: "x"})
--- a/tests/ppx/react/expected/spreadProps.res.txt
+++ b/tests/ppx/react/expected/spreadProps.res.txt
@@ -19,7 +19,7 @@ let c2 = React.createElement(A.make, {...p, x: "x"})
 // let c0 = <A x="x" {...p0} {...p1} />
 
 // only spread props
-let c1 = React.jsx(A.make, p)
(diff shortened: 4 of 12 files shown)


cristianoc (commented):
- add a line to the changelog
- how about `JSXV4.md`, does it need updating?
- how about adding an example with `?key`

tests/ppx/react/expected/externalWithCustomName.res.txt
```diff
@@ -36,4 +36,4 @@ module Foo = {
   external component: React.componentLike<props<int, string>, React.element> = "component"
 }
 
-let t = React.jsx(Foo.component, {a: 1, b: "1"})
+let t = React.jsxWithKey(Foo.component, {a: 1, b: "1"})
```
cristianoc, line 39: why `jsxWithKey` if the key is not passed?

tests/ppx/react/expected/forwardRef.res.txt
```diff
@@ -167,7 +167,10 @@ module V4A = {
       "div",
       {
         children: ?ReactDOM.someElement(
-          React.jsx(FancyInput.make, {ref: input, children: {React.string("Click to focus")}}),
+          React.jsxWithKey(
```
cristianoc, line 170: why jsxWithKey if the key is not passed?

tests/ppx/react/expected/spreadProps.res.txt
```diff
@@ -19,7 +19,7 @@ let c2 = React.createElement(A.make, {...p, x: "x"})
 // let c0 = <A x="x" {...p0} {...p1} />
 
 // only spread props
-let c1 = React.jsx(A.make, p)
+let c1 = React.jsxWithKey(A.make, p)
```
cristianoc, line 22: why `jsxWithKey` is the key is not passed?

tests/ppx/react/expected/spreadProps.res.txt
```diff
@@ -19,7 +19,7 @@ let c2 = React.createElement(A.make, {...p, x: "x"})
...
 
 // reversed order
-let c2 = React.jsx(A.make, {...p, x: "x"})
+let c2 = React.jsxWithKey(A.make, {...p, x: "x"})
```
cristianoc, line 25: why `jsxWithKey` is the key is not passed?

tests/ppx/react/expected/externalWithCustomName.res.txt
mattdamon108: https://github.com/rescript-lang/rescript-react/blob/5e09771ae19b52bdb50061da15265b1093b6b61b/src/React.res#L51:L55
I think I should rename this function. 🤔

mattdamon108:
@cristianoc I think I'm done with this PR including:
* making key type an optional arg of React binding for both classic and automatic
* updating v4 spec with example

cristianoc (commented):
Thanks.
- Would you add an example with `<Foo ?key/>`
- Update the changelog

tests/ppx/react/expected/spreadProps.res.txt
```diff
@@ -23,3 +33,13 @@ let c1 = React.jsx(A.make, p)
 
 // reversed order
 let c2 = React.jsx(A.make, {...p, x: "x"})
+
+let c3 = ReactDOM.jsx("div", p)
```
cristianoc, line 37: There's now something called `jsxNotKeyed` in rescript-react, which never seems to be called. Should it be deleted, or should it be called somewhere?

cli/JSXV4.md
```diff
@@ -286,7 +286,11 @@ React.createElement(Comp.make, {x, y: 7, ?z})
 
 <Comp x key="7">
 // is transformed to
-React.createElement(Comp.make, Jsx.addKeyProp({x: x}, "7"))
+React.createElement(Comp.make, Jsx.addKeyProp(~key="7", {x: x}))
+
+<Comp x key=?Some("7")>
+// is transformed to
+React.createElement(Comp.make, Jsx.addKeyProp(~key=?Some("7"), {x: x}))
```
cknitt, line 293: Isn't it `React.addKeyProp` as per https://github.com/rescript-lang/rescript-react/pull/74?

cknitt:
Also, I was just wondering about the following:

In

```rescript
addKeyProp(~key="42", {a: a, b: b /* and maybe many more props */})
```

*we* are creating the props object, it is not passed from outside. So why do we need to use `Obj.assign` to create a copy of it instead of just mutating it?

mattdamon108:
> * Would you add an example with `<Foo ?key/>`

Do you mean examples in the spec? or test?

cristianoc:
> > * Would you add an example with `<Foo ?key/>`
> 
> Do you mean examples in the spec? or test?

test

tests/ppx/react/expected/spreadProps.res.txt
mattdamon108: Not sure I understand correctly.
`ReactDOM.jsx` calls both `jsxKeyed` and `jsxNotKeyed` https://github.com/rescript-lang/rescript-react/blob/9b0fe1bc548b89ad5b0ddc3acc5bfe85da1630da/src/ReactDOM.res#L1117:L1121

cli/JSXV4.md
mattdamon108: Good catch! I'll fix it.

tests/ppx/react/expected/spreadProps.res.txt
cristianoc: Got it, thanks.

mattdamon108:
> Also, I was just wondering about the following:
> 
> In
> 
> ```rescript
> addKeyProp(~key="42", {a: a, b: b /* and maybe many more props */})
> ```
> 
> _we_ are creating the props object, it is not passed from outside. So why do we need to use `Obj.assign` to create a copy of it instead of just mutating it?

Not sure I understand what you mean.
I think using `Obj.assign` inline would be a little verbose and more tricky to annotate type to put some magic on it. The props type doesn't have a key field, but it has to check the key type and manipulate it.

cristianoc (approved):
Looks good to me.
@mattdamon108 does everything work with your project?

cristianoc:
> Also, I was just wondering about the following:
> 
> In
> 
> ```rescript
> addKeyProp(~key="42", {a: a, b: b /* and maybe many more props */})
> ```
> 
> _we_ are creating the props object, it is not passed from outside. So why do we need to use `Obj.assign` to create a copy of it instead of just mutating it?

The mechanism used does 2 things:
- type checks the normal props and generates the record value for them
- gets around the type checker for `key`, so it does not have to be part of the user-visible prop type definition, and extends the record in the generated code

Even though possible in principle, I'm not sure that both are possible at the same time currently, short of creating some new primitive with special runtime support.

cristianoc:
Oh maybe the question is only about imperatively adding the `key` property. That should be OK.

cknitt:
Yes, I meant that `Obj.assign` is unnecessarily copying the props object, and that we could just do

```rescript
(props->Obj.magic)["key"] = key
```

instead.

mattdamon108:
> Looks good to me. @mattdamon108 does everything work with your project?

Now I tested v3 and v4 classic & automatic, it works fine!

mattdamon108:
> Yes, I meant that `Obj.assign` is unnecessarily copying the props object, and that we could just do
> 
> ```rescript
> (props->Obj.magic)["key"] = key
> ```
> 
> instead.

Maybe this? https://github.com/rescript-lang/rescript-react/pull/74/commits/2f48356f83a11034b748d2d6b16a378aee600808

mattdamon108:
> > Yes, I meant that `Obj.assign` is unnecessarily copying the props object, and that we could just do
> > ```rescript
> > (props->Obj.magic)["key"] = key
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead.
> 
> Maybe this? [rescript-lang/rescript-react@2f48356](https://github.com/rescript-lang/rescript-react/commit/2f48356f83a11034b748d2d6b16a378aee600808)

It also works fine on tests.

mattdamon108:
One impressive thing is that I don't need to change the existing codes between v3 and v4 classic and automatic, except for the Context provider component in my company code base which is a quite large project.
https://github.com/rescript-lang/syntax/blob/master/cli/JSXV4.md#reactcontext

cknitt:
> > > Yes, I meant that `Obj.assign` is unnecessarily copying the props object, and that we could just do
> > > ```rescript
> > > (props->Obj.magic)["key"] = key
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > instead.
> > 
> > 
> > Maybe this? [rescript-lang/rescript-react@2f48356](https://github.com/rescript-lang/rescript-react/commit/2f48356f83a11034b748d2d6b16a378aee600808)
> 
> It also works fine on tests.

We need to ensure that we do not mutate objects passed from user code. So if we do the mutation in `createElementWithKey` and `createElementVariadicWithKey` then we should document that these functions are for use by the PPX only and not meant to be called from user code.

Also, what happens in the case

```rescript
<A {...props} key />
```

? Is props copied before being mutated or just passed through?

cknitt:
Are there no tests for `createElementVariadicWithKey`?

cristianoc:
There's this too https://reactjs.org/docs/react-api.html#cloneelement

mattdamon108:
> > > > Yes, I meant that `Obj.assign` is unnecessarily copying the props object, and that we could just do
> > > > ```rescript
> > > > (props->Obj.magic)["key"] = key
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > instead.
> > > 
> > > 
> > > Maybe this? [rescript-lang/rescript-react@2f48356](https://github.com/rescript-lang/rescript-react/commit/2f48356f83a11034b748d2d6b16a378aee600808)
> > 
> > 
> > It also works fine on tests.
> 
> We need to ensure that we do not mutate objects passed from user code. So if we do the mutation in `createElementWithKey` and `createElementVariadicWithKey` then we should document that these functions are for use by the PPX only and not meant to be called from user code.
> 
> Also, what happens in the case
> 
> ```rescript
> <A {...props} key />
> ```
> 
> ? Is props copied before being mutated or just passed through?

This is a very good point. `<A {...props} key />` would manipulate the usrs' props object.

mattdamon108:
@cknitt Can you move this topic to rescript-react? https://github.com/rescript-lang/rescript-react/pull/74

PR #699, opened 2022-10-25
JSX v4 shared props

mattdamon108:
This PR adds a new feature for JSX v4 which is using shared props type as an argument of `@react.component`
https://github.com/rescript-lang/syntax/issues/695

Example
```rescript
type sharedProps<'a> = {x: 'a, y: string}

// original
module V4C2 = {
  @react.component(:sharedProps<'a>) // <- here
  let make = (~x, ~y) => React.string(x ++ y)
}

// converted to
module V4C2 = {
  type props<'a> = sharedProps<'a>

  @react.component(: sharedProps<'a>) let make = ({x, y, _}: props<'a>) => React.string(x ++ y)
  let make = {
    let \"SharedProps$V4C2" = (props: props<_>) => make(props)

    \"SharedProps$V4C2"
  }
}
```

PR #701, opened 2022-10-26
Update v4 spec about sharing props type

mattdamon108:
This PR updates the JSX v4 spec about sharing props type regarding the PR #699

### DIFF
--- a/cli/JSXV4.md
+++ b/cli/JSXV4.md
@@ -379,30 +379,27 @@ let p: A.props<_> = {x: "x", y: "y"}
 V4 introduces support to set a props type from the outside. It allows sharing of the props type between components.
 
 ```rescript
-type sharedProps<'a> = {x: 'a, y: string}
+type sp = {x: string, y: string}
+type sp1<'a> = {x: 'a, y: string}
+type sp2<'a, 'b> = {x: 'a, y: 'b}
 
-module A = {
-  @react.component(:sharedProps<'a>)
-  let make = (~x, ~y) => React.string(x ++ y)
+module C1 = {
+  @react.component(:sp)
+  let make = (~x, ~y) => body
 }
 
-module B = {
-  @react.component(:sharedProps<'a>)
-  let make = (~x, ~y) => React.string(x ++ y)
+module C2 = {
+  @react.component(:sp1<'a>)
+  let make = (~x, ~y) => body
 }
 
-// is transformed to
-module A = {
-  type props<'a> = sharedProps<'a>
-
-  let make = ({x, y, _}: props<'a>) => React.string(x ++ y)
-  ...
+module C3 = {
+  @react.component(:sp2<'a, 'b>)
+  let make = (~x, ~y) => body
 }
 
-module B = {
-  type props<'a> = sharedProps<'a>
-
-  let make = ({x, y, _}: props<'a>) => React.string(x ++ y)
-  ...
+module C4 = {
+  @react.component(:sp)
+  let make = (~x:int, ~y) => body // type annotation is ignored by type sp1
 }
 ```
--- a/cli/JSXV4.md
+++ b/cli/JSXV4.md
@@ -407,12 +407,12 @@ type sharedprops<'x, 'y> = {x: 'x, y: 'y, z: string}
 
 module C1 = {
   type props<'a, 'b> = sharedProps<'a, 'b>
-  let make = ({x, y, _}: props<'a, 'b>) => React.string(x ++ y ++ z)
+  let make = ({x, y, _}: props<_>) => React.string(x ++ y ++ z)
 }
 
 module C2 = {
   type props<'b> = sharedProps<string, 'b>
-  let make = ({x, y, _}: props<'b>) => React.string(x ++ y ++ z)
+  let make = ({x, y, _}: props<_>) => React.string(x ++ y ++ z)
 }
 
 module C3 = {
--- a/cli/JSXV4.md
+++ b/cli/JSXV4.md
@@ -383,23 +383,23 @@ type sp = {x: string, y: string}
 type sp1<'a> = {x: 'a, y: string}
 type sp2<'a, 'b> = {x: 'a, y: 'b}
 
-module C1 = {
+module C = {
   @react.component(:sp)
   let make = (~x, ~y) => body
 }
 
-module C2 = {
+module C1 = {
   @react.component(:sp1<'a>)
   let make = (~x, ~y) => body
 }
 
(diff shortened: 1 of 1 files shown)


cli/JSXV4.md
```diff
@@ -373,3 +373,36 @@ let p: A.props<_> = {x: "x", y: "y"}
...
+
+### Shared props type (new feature)
+
+V4 introduces support to set a props type from the outside. It allows sharing of the props type between components.
+
+```rescript
+type sharedProps<'a> = {x: 'a, y: string}
+
+module A = {
+  @react.component(:sharedProps<'a>)
```
cristianoc, line 385: Would you clarify in the text above what's the role of `<'a>`. Does the name `'a` matter, or is it only used to specify the number of type arguments?

cristianoc: Not sure I understand: or can be 1 zero type args, or 2 more than zero. And the user needs to specify which one is true?

mattdamon108: Sorry for making communication cost. My point was that I expected there were some use cases:
```rescript
type sp1 = ...

@react.component(:sp1)
```
```rescript
type sp2<'a> = ...

@react.component(:sp2<'a>)
// or
@react.component(:sp2<string>)
```
```rescript
type sp3<'a, 'b, ...> = ...

@react.component(:sp3<'a, 'b, ...>)
// or
@react.component(:sp3<string, 'b, ...>)
```
The text is saying that the type argument(s) is available, the name of `'a` wouldn't matter.

cristianoc: > Sorry for making communication cost. My point was that I expected there were some use cases:
> 
> ```rescript
> type sp1 = ...
> 
> @react.component(:sp1)
> ```
> 
>  ```rescript
> type sp2<'a> = ...
> 
> @react.component(:sp2<'a>)
> // or
> @react.component(:sp2<string>)
> ```
> 
>  ```rescript
> type sp3<'a, 'b, ...> = ...
> 
> @react.component(:sp3<'a, 'b, ...>)
> // or
> @react.component(:sp3<string, 'b, ...>)
> ```
> 
> The text is saying that the type argument(s) is available. The name of `'a` nor the number of type arguments wouldn't matter.

The name does not matter. Does the number matter?

cristianoc: Does it work also if the type definition has zero type arguments?

mattdamon108: If you mean this case:
```rescript
type sp1 = ...

@react.component(:sp1)
```
Yes, it works. https://github.com/rescript-lang/syntax/blob/master/tests/ppx/react/expected/sharedProps.res.txt#L38

cristianoc: I read the source code. So the number of type variables does matter. The name matters only so much as it normally matters, but not less.

So I think what's missing in the doc is one example with 0 and one 2 type variables.

cristianoc: One should also remark on the fact that a type called `prop` cannot be defined in the current module or there will be a type error due to type redefinition. (It could at most be defined outside the current module)

mattdamon108: > One should also remark on the fact that a type called `props` cannot be defined in the current module or there will be a type error due to type redefinition. (It could at most be defined outside the current module)

I think this is already noted as https://github.com/rescript-lang/syntax/blob/master/cli/JSXV4.md#transformation-for-component-definition

mattdamon108: How does it look? https://github.com/rescript-lang/syntax/pull/701/commits/c052cb99bf5a0b99a247b929a3bf170635ca0dc8

cristianoc:
@mattdamon108 made some changes: can you take a look?

What's not clear to me is why one does this:

```res
  let make = ({x, y, _}: props<'a, 'b>) => React.string(x ++ y ++ z)
```

and not this:
```res
  let make = ({x, y, _}: props<_>) => React.string(x ++ y ++ z)
```

cristianoc:
Also what's unclear to me is why the feature combines a partial type definition (passing certain arguments), which can e.g. use a wrong number of arguments or something. And what's the type error going to look like if used wrong.

Another possibility is having the simples mechanism possible, and just pass the name of the type, and provide zero additional type definitions in the PPX. And let the user fully define that type. Probably requiring that it has at least one type argument (in the instructions for use), so there's no special case to worry about.

mattdamon108:
> @mattdamon108 made some changes: can you take a look?
> 
> What's not clear to me is why one does this:
> 
> ```rescript
>   let make = ({x, y, _}: props<'a, 'b>) => React.string(x ++ y ++ z)
> ```
> 
> and not this:
> 
> ```rescript
>   let make = ({x, y, _}: props<_>) => React.string(x ++ y ++ z)
> ```

Actually, the feature generates the latter. https://github.com/rescript-lang/syntax/blob/master/tests/ppx/react/expected/sharedProps.res.txt

mattdamon108:
> Also what's unclear to me is why the feature combines a partial type definition (passing certain arguments), which can e.g. be used with a wrong number of arguments or something. And what's the type error going to look like if used wrong.

Here is the example to show how it works and my logic:

Case 1. simple case
```rescript
type sp = ...

@react.component(:sp)
// transformed to
type props = sp
```

Case 2.
```rescript
type sp1<'a> = ...

@react.component(:sp1<'a>) // Ptyp_var
// transformed to
type props<'a> = sp<'a> // props<'a>
```

Case 3.
```rescript
type sp1<'a> = ...

@react.component(:sp<string>) // Ptyp_constr
// transformed to
type props = sp<string> // just props
```

I just show the case of having 1 type argument example, but it would be the same for the 2, 3, ... cases. Both cases 2 and 3 are possible as use cases. So, if the core_type in payload of `@react.component` has type arguments, then I distinguish the arguments between `Ptyp_var` or else. This is why it makes the partial type definition.

> Another possibility is having the simples mechanism possible, and just pass the name of the type, and provide zero additional type definitions in the PPX. And let the user fully define that type. Probably requiring that it has at least one type argument (in the instructions for use), so there's no special case to worry about.

Yes, I agree with your point. That would make more clean implementation as well as a clear user guide.
IMHO, I guess the current logic would not make many problems in the userland. If the user encounters the type error caused by a mismatched number of type arguments, then would the user rather define the type fully?

cristianoc:
I think the tradeoffs in terms of implementations and uses are clear. Up to you to choose which one looks best.

mattdamon108:
I've made a PR to fix and update the spec accordingly.

mattdamon108:
> Ready to merge?

Yes!

PR #709, opened 2022-10-28
Change char payload

butterunderflow:
Cooperative PR with https://github.com/rescript-lang/rescript-compiler/pull/5759


compiler-libs-406/parmatch.ml
```diff
@@ -376,7 +376,7 @@ let is_cons = function
 
 let pretty_const c = match c with
 | Const_int i -> Printf.sprintf "%d" i
-| Const_char c -> Printf.sprintf "%C" c
+| Const_char i -> Printf.sprintf "%C" (Char.unsafe_chr i)
 | Const_string (s, _) -> Printf.sprintf "%S" s
```
bobzhang, line 380: This function should be adapted, since our char could be a large value. The '%c' may not be appropriate

compiler-libs-406/pprintast.ml
```diff
@@ -192,7 +192,7 @@ let rec longident f = function
 let longident_loc f x = pp f "%a" longident x.txt
 
 let constant f = function
-  | Pconst_char i -> pp f "%C"  i
+  | Pconst_char i -> pp f "%C" (Char.unsafe_chr i) (* todo: consider safety *)
   | Pconst_string (i, None) -> pp f "%S" i
```
bobzhang, line 196: same as above

src/res_printer.ml
```diff
@@ -553,18 +553,20 @@ let printConstant ?(templateLiteral = false) c =
...
-      | '\n' -> "\\n"
-      | '\t' -> "\\t"
-      | '\r' -> "\\r"
-      | '\b' -> "\\b"
-      | ' ' .. '~' as c ->
-        let s = (Bytes.create [@doesNotRaise]) 1 in
-        Bytes.unsafe_set s 0 c;
-        Bytes.unsafe_to_string s
-      | c -> Res_utf8.encodeCodePoint (Obj.magic c)
+      if c <= 127 then
```
bobzhang, line 556: why `if c <= 127` needed?

src/res_scanner.ml
```diff
@@ -508,14 +507,13 @@ let scanEscape scanner =
         | '}' -> next scanner
         | _ -> ());
         let c = !x in
-        if Res_utf8.isValidCodePoint c then Char.unsafe_chr c
-        else Char.unsafe_chr Res_utf8.repl
+        c
```
bobzhang, line 510: why such simplification?

butterunderflow: This is wrong, will fix it.

src/res_printer.ml
butterunderflow: I tried to avoid the usage of Char.unsafe_chr. 
Maybe the following code is clearer?
```ml
  let str = 
        match Char.unsafe_chr c with
        | '\'' -> "\\'"
        | '\\' -> "\\\\"
        | '\n' -> "\\n"
        | '\t' -> "\\t"
        | '\r' -> "\\r"
        | '\b' -> "\\b"
        | ' ' .. '~' as c ->
          let s = (Bytes.create [@doesNotRaise]) 1 in
          Bytes.unsafe_set s 0 c;
          Bytes.unsafe_to_string s
        | _ -> Res_utf8.encodeCodePoint c
```

compiler-libs-406/parmatch.ml
butterunderflow: Sorry, I got stuck here, do you have any suggestion to safely print `int` as `char`?

src/res_printer.ml
```diff
@@ -553,7 +553,7 @@ let printConstant ?(templateLiteral = false) c =
   | Pconst_float (s, _) -> Doc.text s
   | Pconst_char c ->
     let str =
-      match c with
+      match Char.unsafe_chr c with
```
butterunderflow, line 556: There's still a `unsafe_char` usage, but I think this one is safe because it's just value testing and no I/O is involved.

bobzhang: It seems okay, but there may be some duplication with string_of_int_as_char, you may do a clean up later after this commit landed

bobzhang (commented):
looks good to me

PR #711, opened 2022-10-29
Treat await as an almost-unary operator.

cristianoc:
Making pipe behave just like application seems pretty unlikely. First `->` has deep binary operator behavior, including associating to the left on `a->b->c`, and mixes in specific ways with other operators. Instead application lives on a different level entirely with no issues of how to treat specially the right-hand-side which is enclosed in parens (and lives in a radically different place in the call stack of the parser).

The await operator is essentially a unary operator. After all it takes one argument. It is also treated as such in JS. Some expected behaviours are that `await 2 + await 3` just like `-2 + -3` is expected to mean `(await 2) + (await 3)`. Or: `await getName() ++ await getSurname()`.

The issue raised by people who have tried `await` is in combination with pipe: the desired outcome is that `await a->b` is parsed as `await (a->b)`. However notice that `- a->b` is parsed as `(-a)->b`, so that's a difference.

So on one side, one would like to have `await` behave like other unary operators. On the other side one wouldn't.

As a compromise, this PR treats `await` as an almost-unary operator. In the sense that it binds more tightly than any "real" binary operator such as `+` and `**`, but not `->` and `.`.

This introduces some lack of uniformity, and requires special-casing parts of parsing and printing.

### DIFF
--- a/src/res_parens.ml
+++ b/src/res_parens.ml
@@ -175,7 +175,7 @@ let flattenOperandRhs parentOperator rhs =
   | _ when ParsetreeViewer.isTernaryExpr rhs -> true
   | _ -> false
 
-let lazyOrAssertOrAwaitExprRhs expr =
+let lazyOrAssertOrAwaitExprRhs ?(inAwait = false) expr =
   let optBraces, _ = ParsetreeViewer.processBracesAttr expr in
   match optBraces with
   | Some ({Location.loc = bracesLoc}, _) -> Braced bracesLoc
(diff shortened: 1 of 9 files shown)


cristianoc:
@cknitt this should be ready to go. Take a look: it should address the last comment on printing.

src/res_parens.ml
```diff
@@ -175,7 +175,11 @@ let flattenOperandRhs parentOperator rhs =
   | _ when ParsetreeViewer.isTernaryExpr rhs -> true
   | _ -> false
 
-let lazyOrAssertOrAwaitExprRhs expr =
+let binaryOperatorInsideAwaitNeedsParens operator =
+  ParsetreeViewer.operatorPrecedence operator
```
mattdamon108, line 179: Is there a behind history? it seems different from `Res_token.precedence`.

mattdamon108 (approved):

cknitt (approved):
Works fine now! Thanks a lot!

PR #719, opened 2022-11-06
Fix JSX v4 spec about configuration

mattdamon108:
This PR fixes the v3 compatible mode configuration example.

### DIFF
--- a/cli/JSXV4.md
+++ b/cli/JSXV4.md
@@ -22,9 +22,10 @@ To build certain dependencies in V3 compatibility mode, whatever the version use
 
 ```json
 "jsx": {
- "version": 4,
- "v3-dependencies": ["rescript-react-native", "rescript-react-navigation"]
-}
+  "version": 3,
+  "v3-dependencies": ["rescript-react-native", "rescript-react-navigation"]
+},
+"bsc-flags": ["-open ReactV3"]
 ```
 
 In V3 compatibility mode, the listed dependencies are built in V3 mode, and in addition `-open ReatcV3` is added to the compiler options, so that the `ReactV3` compatibility module in rescript-react is used.
@@ -378,7 +379,6 @@ let p: A.props<_> = {x: "x", y: "y"}
 
 V4 introduces support to control the definition of the `props` type by passing as argument to `@react.component` the body of the type definition of `props`. The main application is sharing a single type definition across several components. Here are a few examples:
 
-
 ```rescript
 type sharedprops<'x, 'y> = {x: 'x, y: 'y, z:string}
 
@@ -400,6 +400,7 @@ module C3 = {
 ```
 
 The generated code (some details removed) looks like this:
+
 ```rescript
 @@jsxConfig({version: 4, mode: "classic"})
 
--- a/cli/JSXV4.md
+++ b/cli/JSXV4.md
@@ -18,7 +18,9 @@ To build an entire project in V4 mode, including all its dependencies, use the n
 ### Dependency-level config
 
 Dependencies inherit the `jsx` configuration of the root project. So if the root project uses V4 then the dependencies are built using V4, and the same for V3.
-To build certain dependencies in V3 compatibility mode, whatever the version used in the root project, use `"v3-dependencies"` as in the example:
+To build certain dependencies in V3 compatibility mode, whatever the version used in the root project, use `"v3-dependencies"`: the listed dependencies will be built in V3 mode, and in addition `-open ReatcV3` is added to the compiler options.
+
+For example, suppose a V3 project uses rescript-react 0.11, which requires compatibility mode if compiled with V3, and that 2 dependencies `"rescript-react-native", "rescript-react-navigation"` only build with compatibility more. Then the setting will be:
 
 ```json
 "jsx": {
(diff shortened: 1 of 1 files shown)


cli/JSXV4.md
```diff
@@ -22,9 +22,10 @@ To build certain dependencies in V3 compatibility mode, whatever the version use
 
 ```json
 "jsx": {
- "version": 4,
- "v3-dependencies": ["rescript-react-native", "rescript-react-navigation"]
-}
+  "version": 3,
+  "v3-dependencies": ["rescript-react-native", "rescript-react-navigation"]
```
cristianoc, line 26: This flag specifies on what dependencies v3 is activated. It's not useful if the default version is 3.

You might be thinking of a different examples, where everything builds on v3?

cristianoc: Actually it does also add `-open ReactV3`.

There seem so the 2 use cases:
- Take an existing project and build it with latest rescript-react. Question is: why would you do that if the old rescript-react works and this is the only change. But it can be a stepping stone.
- Migrate a project towards V4 without converting all the dependencies to V3 first.

mattdamon108: I think we have to look into it because I have something strange. Here is the repro example.
* The root project with rc-3 version of compiler and rescript-react. The jsx version config in bsconfig is 3.
* The root project has dependency which is not compatible to v4, e.g. rescript-relay@1.0.2

The rescript-relay is supposed to be built as v3, but it wouldn't. It throws the error:
```
FAILED: src/RescriptRelay.cmj

  We've found a bug for you!
  /Users/woonki/Github/works/gl/sources/sinsunhi-frontend-mirror/node_modules/rescript-relay/src/RescriptRelay.res:666:9-75

  664 ┆   React.createElement(
  665 ┆     provider,
  666 ┆     {"value": Some({"environment": environment}), "children": children
      ┆ },
  667 ┆   )
  668 ┆ }

  This has type:
    {"children": 'a, "value": option<{"environment": Environment.t}>}
  Somewhere wanted: React.Context.props<option<contextShape>>
```
If I add the `v3-dependencies: ["rescript-relay"]`, the error is gone.

Maybe, the jsx config in the root project is not propagated to the dependency which has the old config `"reason": {"react": 3}` until it is added to `v3-dependencies`, I guess.

mattdamon108: We had a same discussion before https://github.com/rescript-lang/rescript-react/pull/49#issuecomment-1253530706

cristianoc: Updated. How does it look?

cli/JSXV4.md
```diff
@@ -18,16 +18,26 @@ To build an entire project in V4 mode, including all its dependencies, use the n
 ### Dependency-level config
 
 Dependencies inherit the `jsx` configuration of the root project. So if the root project uses V4 then the dependencies are built using V4, and the same for V3.
-To build certain dependencies in V3 compatibility mode, whatever the version used in the root project, use `"v3-dependencies"` as in the example:
+To build certain dependencies in V3 compatibility mode, whatever the version used in the root project, use `"v3-dependencies"`: the listed dependencies will be built in V3 mode, and in addition `-open ReatcV3` is added to the compiler options.
+
+For example, suppose a V3 project uses rescript-react 0.11, which requires compatibility mode if compiled with V3, and that 2 dependencies `"rescript-react-native", "rescript-react-navigation"` only build with compatibility more. Then the setting will be:
```
mattdamon108, line 23:
```suggestion
For example, suppose a V3 project uses rescript-react 0.11, which requires compatibility mode if compiled with V3, and that 2 dependencies `"rescript-react-native", "rescript-react-navigation"` only build with compatibility mode. Then the setting will be:
```
Maybe typo.

cristianoc:
@mattdamon108 
Other than the typo, how does it read -- is it clear enough?

cli/JSXV4.md
mattdamon108: Thanks! It looks much better.
Still wondering how much the second use case is useful though. Because the root project could not use the component from the v3 dependencies which doesn’t have a props type.
But it is stating more informative example and explanation now!

mattdamon108:
> @mattdamon108
> Other than the typo, how does it read -- is it clear enough?

Yes! It looks clear now. Thanks!

cli/JSXV4.md
```diff
@@ -18,16 +18,26 @@ To build an entire project in V4 mode, including all its dependencies, use the n
 ### Dependency-level config
 
 Dependencies inherit the `jsx` configuration of the root project. So if the root project uses V4 then the dependencies are built using V4, and the same for V3.
```
mattdamon108, line 20: Is it still correct?

cristianoc: Yes

mattdamon108: I confirm that it is correct. I found that the error was caused by `bsc-flags` with `-open ReactV3`missing.

mattdamon108: I considered adding `"-open ReactV3"` into `bsc-flags` automatically under configuration `jsx.version=3` for seconds, but the premise is needed that the latest rescript-react should be one of the dependencies.
Yeap, it is too much.

mattdamon108: I'm clear now.

cristianoc:
Need to move this to the compiler repo when the syntax repo changes are done.
I guess it does not matter having this on 10.1 and master is fine.

PR #683, opened 2022-10-15
Simplify parsing and printing attributes of function and arguments 

mattdamon108:
Regarding issue https://github.com/rescript-lang/rescript-compiler/issues/5735

This PR improves the parsing and printing attributes of functions and arguments so that it makes the implementation simple and clear for future development extensibility.

PR #722, opened 2022-12-14
Simplify parsing attributes of function and args

mununki:
This PR revives the closed PR https://github.com/rescript-lang/syntax/pull/683 It also makes sync to the compiler. https://github.com/rescript-lang/rescript-compiler/pull/5783


cristianoc (approved):
Thanks for reviving this.

PR #723, opened 2022-12-14
Fix dropping attributes of props in JSX v4

mununki:
This PR is to backport https://github.com/rescript-lang/rescript-compiler/pull/5905


cristianoc (approved):
Thank you!
I guess something should be added to the docs or somewhere to expose this functionality.

mununki:
> Thank you! I guess something should be added to the docs or somewhere to expose this functionality.

Oh! That's a good idea. I'll add it. https://github.com/rescript-association/rescript-lang.org/issues/626

PR #731, opened 2023-02-03
JSX4: Fix missing attributes from props type

mununki:
This PR fixes the issue https://github.com/rescript-lang/rescript-compiler/issues/5969
This change keeps the attributes of the argument in the make function and removes the unnecessary optional attr from the core_type as the type parameter. So it prevents that uncurried functions were incorrectly converting the type of a prop given as a default value.
The current master in the compiler transforms the uncurried type annotation correctly without this change, but this change can fix the issue in v10.1.2

### DIFF
--- a/cli/reactjs_jsx_v4.ml
+++ b/cli/reactjs_jsx_v4.ml
@@ -719,19 +719,10 @@ let argToType ~newtypes ~(typeConstraints : core_type option) types
   in
   match (type_, name, default) with
   | Some type_, name, _ when isOptional name ->
-    ( true,
-      getLabel name,
-      attrs,
-      loc,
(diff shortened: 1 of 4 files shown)


cli/reactjs_jsx_v4.ml
```diff
@@ -719,19 +719,10 @@ let argToType ~newtypes ~(typeConstraints : core_type option) types
   in
   match (type_, name, default) with
   | Some type_, name, _ when isOptional name ->
-    ( true,
-      getLabel name,
-      attrs,
-      loc,
-      {type_ with ptyp_attributes = optionalAttrs} )
```
mununki, line 726: Noted the unnecessary adding of the optional attribute for the core_type.

cristianoc (approved):

PR #732, opened 2023-02-06
Fix issue with printing async functions with locally abstract types

cristianoc:


cristianoc:
This is the syntax part of back-porting this PR: https://github.com/rescript-lang/rescript-compiler/pull/5982

PR #733, opened 2023-02-06
Fix support for recursive components in JSX4

mununki:
This PR fixes the issue https://github.com/rescript-lang/rescript-compiler/issues/5981, backporting https://github.com/rescript-lang/rescript-compiler/pull/5986 to syntax

### DIFF
--- a/cli/reactjs_jsx_v4.ml
+++ b/cli/reactjs_jsx_v4.ml
@@ -1190,14 +1190,15 @@ let transformStructureItem ~config mapper item =
             | Recursive ->
               ( [
                   bindingWrapper
-                    (Exp.let_ ~loc:emptyLoc Recursive
-                       [
-                         makeNewBinding binding expression internalFnName;
-                         Vb.mk
-                           (Pat.var {loc = emptyLoc; txt = fnName})
-                           fullExpression;
-                       ]
-                       (Exp.ident {loc = emptyLoc; txt = Lident fnName}));
+                    (Exp.let_ ~loc:emptyLoc Nonrecursive
+                       [makeNewBinding binding expression internalFnName]
+                       (Exp.let_ ~loc:emptyLoc Nonrecursive
+                          [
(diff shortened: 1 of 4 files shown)


cristianoc (approved):

cli/reactjs_jsx_v4.ml
```diff
@@ -1228,7 +1233,8 @@ let transformStructureItem ~config mapper item =
       List.fold_right otherStructures structuresAndBinding ([], [], [])
     in
     types
-    @ [{pstr_loc; pstr_desc = Pstr_value (recFlag, bindings)}]
+    (* No need to be Recursive as internal representation is called *)
+    @ [{pstr_loc; pstr_desc = Pstr_value (Nonrecursive, bindings)}]
```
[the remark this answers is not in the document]
mununki: It is removed.

cristianoc (approved):
I think it works.

tests/ppx/react/expected/v4.res.txt
```diff
@@ -16,3 +16,55 @@ module AnotherName = {
...
+module Rec = {
+  type props = {}
+
+  let rec make = {
+    @merlin.focus
+    let \"make$Internal" = (_: props) => {
+      make(({}: props))
+    }
+    let make = {
+      let \"V4$Rec" = props => \"make$Internal"(props)
```
cristianoc, line 29: This part seems to be doing noting in the case of recursive functions.
That was true already before, so that's OK. Just, not very useful.

cristianoc: Since recursive components are rare, let's just move on.

mununki: This part seems to be making this expression in js output:
```js
function make(_props) {
  while(true) {
    _props = {};
    continue ;
  };
}

var V4$Rec = { // <- here
  make: make
};
```

PR #734, opened 2023-02-07
Fix issue overlapping argument with default value and alias with default value

mununki:
This PR fixes https://github.com/rescript-lang/rescript-compiler/issues/5977

### DIFF
--- a/tests/ppx/react/expected/aliasProps.res.txt
+++ b/tests/ppx/react/expected/aliasProps.res.txt
@@ -50,7 +50,7 @@ module C2 = {
     | None => ""
     }
 
-    ReactDOM.jsx("div", {})
+    React.string(bar)
   }
   let make = {
     let \"AliasProps$C2" = (props: props<_>) => make(props)
--- a/tests/ppx/react/expected/aliasProps.res.txt
+++ b/tests/ppx/react/expected/aliasProps.res.txt
@@ -9,7 +9,7 @@ module C0 = {
     | Some(text) => text
     | None => "Test"
     }
-    and priority = props.priority
+    and _ = props.priority
 
     React.string(text)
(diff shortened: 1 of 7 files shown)


tests/ppx/react/expected/aliasProps.res.txt
```diff
@@ -4,11 +4,12 @@ module C0 = {
   type props<'priority, 'text> = {priority: 'priority, text?: 'text}
 
   @react.component
-  let make = ({priority: _, ?text, _}: props<'priority, 'text>) => {
-    let text = switch text {
+  let make = (props: props<'priority, 'text>) => {
```
mununki, line 7: Using `props` instead of destructuring to avoid the conflict from overlapped argument name with default value.

mununki:
https://github.com/rescript-lang/syntax/pull/734/commits/74e71b16c7f21e7d2cf4f97b167b8bc019b452e1 fixes the issue https://github.com/rescript-lang/rescript-compiler/issues/5979
These two issues are tightly related, and I think it is better to fix them with this change.

mununki:
> Could you also add tests that do compile, so it's clear what little mistakes are in there. As in `jscomp/test/recursive_react_component.res`.

Do you mean in the compiler repo?

cristianoc:
> > Could you also add tests that do compile, so it's clear what little mistakes are in there. As in `jscomp/test/recursive_react_component.res`.
> 
> Do you mean in the compiler repo?

Yes sorry, I mean, do this PR in the compiler repo first.

mununki:
Backporting https://github.com/rescript-lang/rescript-compiler/pull/5989

cristianoc (approved):
Looking good.

PR #735, opened 2023-02-24
Fix formatting of `switch` expressions that contain braced cases inside

ah-yu:
Backporting https://github.com/rescript-lang/rescript-compiler/pull/6015


cristianoc (approved):
Thanks!

PR #737, opened 2023-03-03
Fix issue with JSX V4 and newtype

cristianoc:
This is a port of PR https://github.com/rescript-lang/rescript-compiler/pull/6029


mununki (approved):
Thanks!

